PR #62, opened 2014-09-23
Update utils.system to return stdout and stderr separately

embray:
I noticed recently that at some point after v3.49 (i.e. now whenever I pip install the latest zest.releaser) I would have trouble with occasional garbage in my version string or project name, since zest.releaser is using `./setup.py --name` and `./setup.py --version` to obtain those`.

However, although those commands should only output the name/version on stdout, our setup.py is such that occasionally some warnings will go to stderr.  Since the `system()` function just concatenates stdout and stderr that leads to problems.  This is a possible problem in some other cases too--for example git sometimes puts warnings stderr that could confuse things when trying to parse what came out of stdout.

This changes `system()` to return a `(stdout, stderr)` tuple.  In cases where it's appropriate to display the output from both streams it's easy enough to do `''.join(system(<cmd>))`.  Alternatively, I could offer a new argument to `system()` to return only stdout.

### DIFF
--- a/zest/releaser/bzr.py
+++ b/zest/releaser/bzr.py
@@ -24,7 +24,7 @@ class Bzr(BaseVersionControl):
         return dir_name
 
     def available_tags(self):
-        tag_info = system('bzr tags')[0]
+        tag_info = system('bzr tags', stderr=False)
         tags = [line[:line.find(' ')] for line in tag_info.split('\n')]
         tags = [tag for tag in tags if tag]
         logger.debug("Available tags: %r", tags)
@@ -71,11 +71,11 @@ class Bzr(BaseVersionControl):
         status' and see what we get.
         """
         # Check for changes to versioned files.
-        if system('bzr status --versioned')[0]:
+        if system('bzr status --versioned', stderr=False):
             # Local changes.
             return False
         return True
 
     def list_files(self):
         """List files in version control."""
-        return system('bzr ls --recursive')[0].splitlines()
+        return system('bzr ls --recursive', stderr=False).splitlines()
--- a/zest/releaser/bzr.py
+++ b/zest/releaser/bzr.py
@@ -24,7 +24,7 @@ class Bzr(BaseVersionControl):
         return dir_name
 
     def available_tags(self):
-        tag_info = system('bzr tags', stderr=False)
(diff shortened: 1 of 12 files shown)


embray:
Nearly forgot about this until I was bitten by it again.  Should I still go ahead and fix up this PR?

reinout:
Well, I can do some work on it too. It'll be friday morning or afternoon (Dutch timezone).

Seeing the large number of places where you had to add the `=False` argument... Wouldn't it make sense to only return the stdout (which is what you want) and, when there's stderr output, don't return it but just do a `logger.debug()` with it. If there's a weird error somewhere, you could run zest.releaser with `--verbose` and see the error message.

I never really looked at what gets passed as stdout and what as stderr... I just copied the `system()` method almost verbatim from zc.buildout.

So... What's your opinion on not returning the stderr at all? Everything *should* keep working, I guess.

I'll put it on my todo list to look at it within 16 hours. Feel free to update the pull request somewhat if should want to try out only returning stdout.

embray:
I think that sounds like a good plan.  Most of the time I don't care about what's on stderr (i.e. warnings) unless there's an actual error (i.e. return code != 0).

reinout:
Ah yes, of course. So IF the exit code is 0 we just return stdout. IF otherwise, we return stdout + stderr as usual.

As an extra, we `logger.debug()` the stderr output even when the exit code is 0.

Something like that?

reinout:
Actually, this might cut down on quite some test errors that are due to weird warnings... Supporting all the different versions of git/bzr/hg/svn/svn2git is a pain :-)

embray:
This--the reason this ends up biting me is weird git warnings (really, I just need to upgrade my git).

reinout:
Ok, I tried it as a quick fix in #69, but that's giving me a load of test errors locally. Let's see what travis tells us. Feel free to copy/paste around from my branch to yours, I'm almost off to bed :-)

reinout:
Closing this pull request as I've noticed that the stderr is necessary in almost all cases (git/bzr/svn and even setup.py). #69 only tries to hide the stderr for `setup.py`. And only if there's no traceback. Let's see if that one works.

For the rest: can you tell us what the output is that trips up zest.releaser? Perhaps we can directly compensate for that.

PR #69, opened 2015-01-15
Only returning stderr if there's actually a non-zero exit code

reinout:
Trying it out as a pull request to get travis to test it.

Locally I get a lot of test errors, but that might just be version differences.

### DIFF
--- a/zest/releaser/svn.py
+++ b/zest/releaser/svn.py
@@ -98,7 +98,7 @@ class Subversion(BaseVersionControl):
         if 'Could not resolve hostname' in tag_info or \
                 'Repository moved' in tag_info or 'E670008' in tag_info:
             logger.error('Network problem: %s', tag_info)
-            sys.exit()
+            sys.exit(1)
         tags = [line.replace('/', '').strip()
                 for line in tag_info.split('\n')]
         tags = [tag for tag in tags if tag]  # filter empty ones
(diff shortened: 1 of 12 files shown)


embray:
Looks good to me at least in principle.

reinout:
Ok, I've finally gotten the tests to run again. Turned out I couldn't ignore stderr for bzr and git and so. Necessary information was hidden that way. And even for calling setup.py, I needed to look at the stderr to see if there's a traceback in there. (As there's not always a non-zero exit code).

So... in the end only non-traceback stderr output of calling setup.py is filtered out.

@embray, does this branch actually still fix your error?

@mauritsvanrees : can you take a look? I think it's OK to merge the pull request as I also made sure the tests run again. And I fixed up some exit codes. But the actual stderr-ignoring code looks a bit like a mess to me.

zest/releaser/svn.py
```diff
@@ -95,10 +97,16 @@ def available_tags(self):
                 sys.exit(0)
 
         tag_info = system('svn list %s%s' % (base, tags_name))
-        if 'Could not resolve hostname' in tag_info or \
-                'Repository moved' in tag_info or 'E670008' in tag_info:
+        network_errors = ['E670008',
+                          'Unable to connect',
+                          'Could not resolve hostname',
+                          'Repository moved',
+                          'E670008']
```
mauritsvanrees: `E670008` is double in this list. Maybe the new `Unable to connect` has a different error code that you wanted to add instead?

mauritsvanrees:
Seems good to me.

embray:
Sorry, kinda vanished for a while.  This should resolve the original issue I raised :+1:

Issue #55, opened 2014-06-04
Handling wheels


hgrecco:
It seems that the wheel format is not handled by zest.releaser. It would be nice that if `[bdist_wheel]` is present in `setup.cfg` then
```
python setup.py sdist bdist_wheel upload
```
is called instead of:
```
python setup.py sdist upload
```

mauritsvanrees:
I created a pull request for this. It simply runs the `bdist_wheel` command at the same time as `sdist` when the `wheel` package is available. It does not check `setup.cfg`. I wonder if that is enough initially, or if we should really check if the package has a `setup.cfg` with a `bdist_wheel` section, which could be empty.

The advantage of the simple approach in this pull request is that packages need no changes.

Can we think of existing packages where this is a bad idea?
I am not sure how smart bdist_wheel is about autodiscovering that a correct and sane wheel is created.
Might be safer to check for a `bdist_wheel` entry in setup.cfg. Or possibly add an option for the zest.releaser section.

reinout:
@mauritsvanrees, you started with a `zest.releaser[recommended]` list that includes 'wheel' right now. See #49 where you talk about that recommended list.

mauritsvanrees:
Is on master and will end up in next release (4.0).

PR #78, opened 2015-04-07
Add optional support for uploading Python wheels.

mauritsvanrees:
Use the new zest.releaser[recommended] extra, or pip install
wheel yourself next to zest.releaser.

Issue #55

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,15 @@ Changelog for zest.releaser
 3.57 (unreleased)
 -----------------
 
+- Add optional support for uploading Python wheels.  Use the new
+  ``zest.releaser[recommended]`` extra, or pip install ``wheel``
+  yourself next to ``zest.releaser``.  Create or edit ``setup.cfg`` in
+  your project (or globally in your ``~/.pypirc``) and create a section
+  ``[zest.releaser]`` with ``create_wheel = yes`` to create a wheel to
+  upload to PyPI.
+  Issue #55
+  [maurits]
+
 - Fix a random test failure on Travis CI, by resetting
   ``AUTO_RESPONSE``.
   [maurits]
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,7 +5,7 @@ Changelog for zest.releaser
 -----------------
 
 - Add optional support for uploading Python wheels.  Use the new
-  ``zest.releaser[recommended]`` extra, or pip install ``wheel``
+  ``zest.releaser[recommended]`` extra, or run ``pip install wheel``
   yourself next to ``zest.releaser``.  Create or edit ``setup.cfg`` in
   your project (or globally in your ``~/.pypirc``) and create a section
   ``[zest.releaser]`` with ``create_wheel = yes`` to create a wheel to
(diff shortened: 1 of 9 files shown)


mauritsvanrees:
Note that I rebased this branch on master, to get my recent test fix in.

I did a rebase and a forced push, as advocated in https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request which showed up in a Google search.
I was skeptical of the forced push, but thought I would try it just this once. Indeed it gives problems when you have already checked out the branch somewhere else and you do a pull, giving you conflicts. So a merge would have been better.
If you have already checked out this branch locally, you may want to remove it before pulling.

For clarity: a merge of this pull request to master would be safe.

reinout:
Yeah, I always merge. Rebasing sounds fun, but there shouldn't be a copy of the branch anywhere else. Otherwise you're asking for a medieval treatment :-)

CHANGES.rst
```diff
@@ -4,6 +4,15 @@ Changelog for zest.releaser
 3.57 (unreleased)
 -----------------
 
+- Add optional support for uploading Python wheels.  Use the new
+  ``zest.releaser[recommended]`` extra, or pip install ``wheel``
+  yourself next to ``zest.releaser``.  Create or edit ``setup.cfg`` in
```
reinout: Needs a verb somewhere. Now I apparently should "pip install" myself :-)

reinout:
Change looks good, apart from one thing: shouldn't wheels be on by default (provided that the wheel package is available)? At least, I can imagine it being the default in 2016. Won't changing it then be hard?

I haven't used wheels much, so I don't know much about the drawbacks.

mauritsvanrees:
I have not used wheels at all. (And zc.buildout cannot handle them yet, so I won't be using them much in the near future.) I have my info from http://pythonwheels.com. From what I read there, I distill this for deciding whether wheels are good for a package:

- pure python 2 or python 3 package: no problem

- using 2to3 for Python 3 support: a problem, though I do not see there how you should then handle this, if you should maybe forget about wheels for now.

- Using other ways to support both Python 2 and 3: no problem, you should upload a *universal* wheel.

- Package with optional C extensions: do NOT upload a univeral wheel. non-universal is presumably fine.

- Package with C extensions in general (optional or default): upload wheel only for Mac or Windows.

That sounded a bit tricky for zest.releaser to decide on its own, so I gave the user an explicit way to indicate this for a package.

Maybe we can come up with something better.

doc/source/uploading.rst
```diff
@@ -89,6 +89,19 @@ You can use no/false/off/0 or yes/true/on/1 as answers; upper, lower or mixed
 case are all fine.
 
 
+Uploading wheels
+----------------
+
+First, you should install the ``zest.releaser[recommended]`` extra, or
+``pip install wheel`` yourself next to ``zest.releaser``.  Then create
```
reinout: Same textual change needed.

reinout:
Ok, clear explanation. Sounds good to leave it optional for now

reinout:
Merge at will.

Can you take a quick peek at #82 and perhaps merge that one as well (in case you know a bit more than me about tox.ini).

And afterwards do a release?

PR #79, opened 2015-04-11
Remove code for support of collective.sdist.

mauritsvanrees:
That package was a backport from distutils for Python 2.5 and earlier,
which we do not support.

@reinout, want to have a look to see if my assumptions are wrong and I break something?

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,11 @@ Changelog for zest.releaser
 3.57 (unreleased)
 -----------------
 
+- Remove code for support of collective.sdist.  That package was a
+  backport from distutils for Python 2.5 and earlier, which we do not
+  support.
+  [maurits]
+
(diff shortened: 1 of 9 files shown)


reinout:
Hey, you removed the 'mregister' command. Wasn't that the "maurits-register" command? :-)

Anyway, looks good to me. Lots of lines removed, which is nice.

mauritsvanrees:
mregister was multiple register I guess. And apparently that was only needed when using collective.dist, as that could not (or did not want to) override the existing register command.

reinout:
I just merged master, can you double-check if I didn't break anything?

After that, you can merge to master and start merging #80, ok? :-)

reinout:
Ah, I *knew* I *had* to have broken something!

mauritsvanrees:
Strange error about not being able to get zope.exceptions from PyPI now, on Python 2.6. I have restarted the job.

Issue #26, opened 2012-09-26
Don't continue when CHANGES are 'nothing changed yet'


rvanlaar:
Sometimes I forget to add the changes to the CHANGES.rst file and I find out during a release.

A fix would be to detect whether or not the CHANGES.rst has some changes for this release.

mauritsvanrees:
I have created a pull request for this. When the 'Nothing changed yet' line is in the changelog, between the two first headers, we now print a warning and ask 'Are you sure you want to release'. Default answer is no. When answering 'no', we simply print a hint to use the lasttaglog command. People will want to use their favorite editor when adding a changelog entry, and I did not want to handle that.

mauritsvanrees:
Related pull request is merged.

PR #80, opened 2015-04-11
Warn when no changelog entry has been added.

mauritsvanrees:
When between the last postrelease and a new prerelease a changelog entry
has not been added, '- Nothing changed yet' would still be in there.

Issue #26.

### DIFF
--- a/zest/releaser/prerelease.py
+++ b/zest/releaser/prerelease.py
@@ -8,6 +8,7 @@ import sys
 from zest.releaser import baserelease
 from zest.releaser import utils
 from zest.releaser.utils import system
+from zest.releaser.postrelease import NOTHING_CHANGED_YET
 
 logger = logging.getLogger(__name__)
 
@@ -25,6 +26,9 @@ DATA = {
     'new_version': 'New version (so 1.0 instead of 1.0dev)',
     'history_file': 'Filename of history/changelog file (when found)',
     'history_lines': 'List with all history file lines (when found)',
+    'nothing_changed_yet': (
+        'First line in new changelog section, '
+        'warn when this is still in there before releasing'),
     'original_version': 'Version before prereleasing (e.g. 1.0dev)',
     'commit_msg': 'Message template used when committing',
     'history_header': 'Header template used for 1st history header',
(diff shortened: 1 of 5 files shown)


zest/releaser/prerelease.py
```diff
@@ -127,6 +132,24 @@ def _grab_history(self):
         # TODO: add line number where an extra changelog entry can be
         # inserted.
 
+        # Look for 'Nothing changed yet' under the latest header.  Not
+        # nice if this text ends up in the changelog.  Did nothing happen?
+        start = headings[0]['line']
```
reinout: Can't `headings` be an empty list? So that `headings[0]` gives an index error?

I wonder: can't we just simply grab the first 10 lines of the changelog, join them together and check for the 'nothing changed' string? It is not something that needs to be very strict, right? It *does* need to be fail-safe and failure-proof. A quick join of the first 10 lines would be enough, I guess.

reinout:
Small comment about elaborate code that can be replaced by something simpler.

Otherwise fine.

zest/releaser/prerelease.py
mauritsvanrees: If headings is empty, we have already raised an error in the condition around line 110 ("No detectable version heading in the history file").

mauritsvanrees: If someone has a few extra lines at the top of the history file, ten lines may not be enough. If you increase it to twenty you may get false positives for earlier releases where you forgot this. Not really bad of course.
I could change it, but in this case I thought: we already have the headers so we know exactly where to look, so let's use it. Is that okay or should I still change it?

reinout: Fine in that case.

I'm merging another pull request with master, so wait before merging this one :-)

mauritsvanrees: Yeah, some of these pull requests touch similar code, and of course the CHANGES.rst as well for conflicts. So I'll keep my hands of it for now if you are busy with it.
It was a productive Plone Open Garden in Sorrento last week for zest.releaser. :-)

PR #81, opened 2015-04-11
Add '[ci skip]' to commit messages.

mauritsvanrees:
This avoids running Travis Continuous Integration builds.  See
http://docs.travis-ci.com/user/how-to-skip-a-build/

To activate this, add '[zest.releaser] ci-skip = yes' to the
'setup.cfg' of a package, or your global '~/.pypirc'.

Eric Steele wants something like this for Plone. :-)

Alternatives I can think of:

- Always do this automatically when you have a `.travis.yml` in your repo.

- Do this in a separate package, using the hooks of zest.releaser.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,13 @@ Changelog for zest.releaser
 3.57 (unreleased)
 -----------------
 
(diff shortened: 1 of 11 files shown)


reinout:
Looks useful!

Issue #59, opened 2014-07-26
Support twine for uploading


wosc:
Since ``setup.py upload`` does not support HTTPS, but [twine](https://pypi.python.org/pypi/twine) does, it would be nice if zest.releaser supported that (at least additionally, but probably even _instead of_ the old method).

reinout:
See #49 in case it is handier to make it an optional dependency.

reinout:
@mgedmin, if you're looking at twine, tackling #20 at the same time is probably very easy?

mauritsvanrees:
I am working on this now. Will create a pull request.

I will note one thing. When a package is not registered to PyPI yet, `twine upload` will fail. You must either manually login to PyPI and register the package, or use `python setup.py register` after all.

See https://github.com/pypa/twine/issues/21

reinout:
This will be fixed in 4.0. You could try master if you wish.

PR #82, opened 2015-04-13
Add a tox.ini

mgedmin:
tox.ini is becoming the standard way to run tests on an unfamiliar
Python package.  Bonus points: it lets you rest with multiple Python
versions conveniently, which is *essential* should you wish to add
Python 3 support.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ zest.releaser.egg-info
 htmlcov
 doc/build/
 
+.tox/
--- a/tox.ini
+++ b/tox.ini
@@ -9,4 +9,4 @@ deps =
     z3c.testsetup
     setuptools >= 8
 commands =
-    zope-testrunner --test-path=. --tests-pattern=^tests$ -v -c
+    zope-testrunner --test-path=. --tests-pattern=^tests$ {posargs:-v -c}
--- a/tox.ini
+++ b/tox.ini
@@ -1,6 +1,6 @@
 [tox]
 envlist =
-    py26,py27,pypy
+    py26,py27
 
 [testenv]
 usedevelop = true
--- a/tox.ini
+++ b/tox.ini
@@ -1,0 +1,11 @@
+[tox]
+envlist =
+    py26,py27,pypy
+
+[testenv]
+usedevelop = true
+deps =
+    zope.testrunner
+    z3c.testsetup
+commands =
+    zope-testrunner --test-path=. --tests-pattern=^tests$ -v -c
--- a/tox.ini
+++ b/tox.ini
@@ -7,5 +7,6 @@ usedevelop = true
 deps =
     zope.testrunner
     z3c.testsetup
+    setuptools >= 8
 commands =
     zope-testrunner --test-path=. --tests-pattern=^tests$ -v -c


reinout:
I know that tox exists, but I haven't used it yet. In principle, adding this file is fine. I guess it means that others can run it with tox?

I currently depend on travis to warn me about errors. Especially as zest.releaser needs quite some external tools like git/bzr/hg/git-svn. 

A question about the tox.ini: it looks as if it runs the tests without using buildout? Because it lists some test dependencies? I guess that's because of the not-quite-standard zope testrunner setup? (I might need to modernize the zest.releaser tests, I guess. The current setup is a left-over from 7 years ago or so :-) )

You added just the tox.ini, does that mean that some usage documentation (perhaps just a link) should be added? I guess just a two-line changelog entry with a 'how to use tox' url would be enough.

mauritsvanrees:
I have done a small bit with tox, when looking at the ``zope.tal`` package last year, so I'll have a look.

I get a few errors in the ``postrelease`` part of the tests when running tox:

```
File "/Users/mauritsvanrees/buildout/tools/src/zest.releaser/zest/releaser/tests/postrelease.txt", line 76, in postrelease.txt
Failed example:
    vcs.version
    - '0.1b0'
    ?      -
    + '0.1b'
```

I guess that is due to having a too new or too old setuptools version. Can tox set a specific version? Or is that not the way to go? Maybe tox should run a bootstrap and buildout first?

The above error actually indicates that we may need to do some changes for some setuptools versions, as the ``postrelease`` command does not work nicely, but that would be totally outside the scope of the pull request.

I wonder if with tox it makes sense to add a test_suite to our ``setup.py``.

On master we have *today* added a ``zest.releaser[recommended]`` extra, which has ``wheel`` as dependency. What would be the way to add that in tox? Add the extra to ``[testenv] deps``? If it is as simple as that, I would not mind adding it myself. Otherwise it would be nice to update your pull request accordingly.

mauritsvanrees:
Ah, the postrelease tests fail because last year I updated them to run with newer setuptools, version 8 or higher (commit 357d32f45db88be40381b775bd2fda49f0c9dfd4). zest.releaser still works fine with older versions, but the tests would indeed fail, and in reality it might not always be able to suggest a nice new version number, but that is no big deal.

It *would* still be nice to know how to get tox to use a specific setuptools version. The virtualenv (python 2.7) where I had installed tox was using ``distribute``, so really ancient. I have updated it to setuptools 15.0, but get the same errors. Would it be something to add to ``[testenv]``? So probably ``devs = setuptools >= 8``. You might argue that this should be in our setup.py, but it still works good enough with lower versions, so it is only irritating while running the tests.

Oh, I have updated tox from 1.6.1 to the latest version 1.9.2. Let's see how that goes. That actually helps. Great!

An entry in CHANGES.rst and a few lines in docs/source/developing.rst would still be nice. Those lines should probably contain a link to https://tox.readthedocs.org/ which me and Reinout should read some time. ;-)

mgedmin:
tox is a bit like Travis that you can run on your own laptop before actually committing and pushing stuff for others to see :)

Speaking from personal experience, it's not a good idea to use buildout from tox. E.g. there's `detox`, a tool that runs tox tests in parallel, ideal for modern multi-core CPUs.  That won't work if the tests try to stomp over the same files (like bin/test or .installed.cfg or basically everything else that buildout does).

As for documentation, it didn't occur to me.  I assume you have existing documentation for how to run the tests with buildout etc?  That will continue to work.  I did this PR for people like myself, who're too lazy to read documentation and instead do an `ls`, see a `tox.ini` and say "great! I know this!".

I don't know if you will want to switch to tox as the primary way of running tests.  Tox has advantages (multi-Python-version support, familiar command-line interface), but it also has disadvantages (setup.py install/develop overhead for every test run, ~~no way to run just a single test or add --pdb~~ actually there's a way, I just need to add `{posargs}` to the zope-testrunner command in tox.ini and you'll be able to do `tox -e py27 -- --pdb`).

Re: setuptools dependency: sure, `setuptools >= 8` in `deps` will work.  As for `zest.releaser[extra_something]`, I'm not 100% sure (and that's why I added `z3c.testsetup` explicitly instead of depending on `zest.releaser[test]`).   Some experimentation might be required.

mgedmin:
Uh.  I'm now seeing failures on Python 2.6.  The cause seems to be this difference:
```
      [egg_info]
    - tag_build =
      tag_svn_revision = false
    + tag_build =
      <BLANKLINE>
```
but of course there's [a lot of other (irrelevant) output](https://gist.github.com/mgedmin/cebcea0c775088abd889).

I swear I didn't get those when I pushed the original PR, but now I'm seeing them even if I git reset --hard  to the older commit.

(While testing this I discovered that tests fail if I checkout an older commit and leave the source tree in a detached HEAD state.  Do the tests seriously rely on the state of git metadata?  That doesn't seem like a good idea to me...)

mgedmin:
Travis failed because
> Download error on http://pypi.python.org/simple/zope.testrunner/: [Errno 110] Connection timed out -- Some packages may not be found!

and I cannot hit the Rebuild button because I'm not a committer.

reinout:
I restarted the build and it passes now. I'll merge it.

Note: I don't actively use tox right now, so I'm not giving guarantees on how up-to-date I'll keep the tox.ini. Though, now it is there, I'll probably try using it. I'll want to support python 3 someday and then it can be very handy.

mgedmin:
Uh, in the mean time I forgot what feature or bugfix I wanted to suggest for zest.releaser when I started figuring out how to run its test suite locally...

Ah!  Using twine for uploading to PyPI securely!

reinout:
Twine? Ah, a pull request to fix #59 :-)

mauritsvanrees:
About Python 2.6 tests with tox: I see the same tests fail randomly: `tox -e py26 -- -t pypi.txt` gives this one plus two similar:

```
File "/Users/mauritsvanrees/buildout/tools/src/zest.releaser/zest/releaser/tests/pypi.txt", line 258, in pypi.txt
Failed example:
    setup_config.fix_config()
    + [egg_info]
    + tag_svn_revision = false
    + tag_build =
    + <BLANKLINE>
      [zest.releaser]
      release = no
      <BLANKLINE>
    + <BLANKLINE>
    - [egg_info]
    - tag_build =
    - tag_svn_revision = false
```

Here ConfigParser reads a setup.cfg, we change a few things and let ConfigParser write it back and then the order is sometimes different... On Python2.7 it is fine. And with our bin/test runner created with buildout it always works. So: very strange.

Probably the only way to fix it is to fix the `read` or `write` method of `ConfigParser` in 2.6. I briefly looked at the differences, but did not see what might cause the different order. We may need to let that one pass and not do anything about it.

@mgedmin, you wrote "(While testing this I discovered that tests fail if I checkout an older commit and leave the source tree in a detached HEAD state. Do the tests seriously rely on the state of git metadata? That doesn't seem like a good idea to me..."
I do not see this happening. I can checkout an older commit, for example a72e07a, and run the tox tests just fine, except for the errors above.

mgedmin:
I noticed the "You're in a detached HEAD state" messages inside the test failure diffs, but it's entirely possible those are covered by +ELLIPSIS matching, and the real failure was ConfigParser's nondeterminism.

(Any reason to continue supporting zest.releaser on Python 2.6?)

mauritsvanrees:
I would not much mind dropping support for Python 2.6.  Python 2.7 has been out for almost 5 years now.

I always find it a not so nice reason to drop support just because tests for a version are hard to keep up to date, especially when it is with tox which we don't officially use. The package runs just fine on 2.6 as far as I know. And 2.6 on Travis is happy.

But it *does* suck energy, which may be better spent elsewhere. Like porting to Python 3, which is easier if we drop 2.6 support. :-)

@reinout : would you be okay with that? With all the changes we are working on now, I was already inclined to increase the major version of zest.releaser. That may be a good moment to drop 2.6 support, even though things *should* still work. The docs could say that it *probably* works on 2.6 but that it is not tested anymore.

reinout:
Fine. Also the version bump. Adding wheel support and especially starting with dependencies is a good reason for a bump.

Dropping official 2.6 support is fine. Only reason for 2.6 would be if ancient plone instances would need it.

mgedmin:
Do... do ancient plone instances perform software releases?

(I understand keeping Python 2.6 support for services people deploy on servers running ancient LTS distributions.  I'm less sure about the value of Python 2.6 for development tools people run on their personal workstations.)

mauritsvanrees:
No, Plone instances do not perform software updates. The [core development buildout](https://github.com/plone/buildout.coredev/blob/4.3/core.cfg#L69) does install it, but that is only so that developers can use it locally for releasing packages, if they really want to. I guess most developers install zest.releaser (or for example jarn.makerelease) somewhere else and use that. That is what I do.

Plone 4.3 does still support Python 2.6, so it is probably best to not update the zest.releaser version pinned in versions.cfg for that branch. Dropping 2.6 support in zest.releaser is no problem for Plone.

PR #83, opened 2015-04-14
Clarified log message to say that an sdist/wheel is created in a temp folder

### LINKED ISSUE #61: Shouldn't the egg land in the dist directory if not uploading to PyPI?
Because the egg didn't go to dist folder (plone/src/product.name/dist) I spent a while looking around for why my egg wasn't built, until I realised that I didn't have my plugins installed.

The egg was built of course - it had just landed in somewhere like:

     /tmp/product.name-0.2.3-Prt03G/gitclone/dist

reinout:
Fixes #61

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -35,6 +35,10 @@ Changelog for zest.releaser
 - Fix a random test failure on Travis CI, by resetting ``AUTO_RESPONSE``.
   [maurits]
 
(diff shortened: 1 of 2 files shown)


reinout:
Sorry, I did something wrong when merging master (I was in a branch instead of in master...). I couldn't rebase as I had already pushed commits to github.

PR #89, opened 2015-05-05
Maurits postrelease version check

mauritsvanrees:
Actually: postrelease *and* prerelease. Any part where we call `ask_version`.

Example commit that we want to avoid here:
https://github.com/plone/plone.app.testing/commit/81c54c6adc72ace456b40cda327f0b8d3bbbf55a
(I have reverted that one in https://github.com/plone/plone.app.testing/commit/24663b9b90d6d849045c75303a0cfc4c4dcbf2d3)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,9 @@ Changelog for zest.releaser
 3.57 (unreleased)
 -----------------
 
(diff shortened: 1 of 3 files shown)


reinout:
Great one!

PR #90, opened 2015-05-05
Catch errors and colorize

mauritsvanrees:
This uses `colorama` in one spot in `utils.system` for printing errors in red.

Note that technically we would not need `colorama`. `print('\x1b[31m' + 'This prints in red')` works just fine without it. But then you need to print another control character to go back to the standard color. And you would probably first need to find out whether that standard color is black or white or something else. And it would not work on Windows. So `colorama` seems useful as dependency to me.

It tripped up the tests, where invisible differences were reported: the difference was the control character. I caught that in the normalizer in `test_setup.py`.

Since I expected to change several print lines, I decided to first switch to the `print()` function instead of the statement, to move closer to Python 3.

And importantly, one of the commits makes sure that errors in PyPI uploads are visible. You can try it out by changing the password in your `~/.pypirc` so it is wrong and you cannot upload. Before:

```
...
Showing last few lines...
Creating tar archive
removing 'collective.searchandreplace-4.1' (and everything under it)
running upload
Submitting dist/collective.searchandreplace-4.1.tar.gz to https://testpypi.python.org/pypi
```

After:

```
...
Showing last few lines...
Submitting dist/collective.searchandreplace-4.1.tar.gz to https://testpypi.python.org/pypi
warning: no previously-included files matching '*.pyc' found anywhere in distribution
Upload failed (401): Incorrect password
error: Upload failed (401): Incorrect password
```

And the last three lines are part of stderr, so they are now printed in red.

Note that I made it so that only the `upload` and `register` commands get this treatment for showing any errors. Warnings are also printed, as you can see above. And with git, when we do a checkout of a tag, you will see in red `HEAD is now at 344f718... Preparing release 4.1`. We will have to see in practice if it makes sense to hide some stuff.


setup.py
```diff
@@ -36,7 +36,7 @@ def read(filename):
       zip_safe=False,
       install_requires=[
           'setuptools',
-          # -*- Extra requirements: -*-
+          'colorama',
```
reinout: So... this is where we cross the line towards dependencies :-)

I've long tried to keep zest.releaser working without dependencies, but I guess now is the time to do it. With wheels and so... And there was some new generic library that might be handy to upload stuff to pypi. Etc.

zest/releaser/__init__.py
```diff
@@ -1,4 +1,10 @@
 import pkg_resources
+from colorama import init
 
+# Initialize colorized output.  Set it to reset after each print, so
+# for example a foreground color does not linger into the next print.
+# Note that the colorama docs say you should call `deinit` at exit,
+# but it looks like it already does that itself.
+init(autoreset=True)
```
reinout: Should this `init()` go here? This way we have another package's init stuff Real High Up in our code.

For explicitness' sake I'd prefer to have it in the same place where we set up logging (probably the `main()` in pre/postrelease.py)

reinout: Oh... that assumes we actually set up logging :-) 

Anyway, I'd prefer to have the init in the main() if practical.

reinout:
Quite a job fixing all those print statements! Good one.

Looks good in principle, apart from the init in the __init__ :-)

mauritsvanrees:
I tried moving the `colorama.init()` to the various `main` methods.  But somehow we get totally no output at all then. Very strange:

```
$ bin/test -t prerelease.txt
Running tests at level 1
Running zope.testrunner.layer.UnitTests tests:
  Set up zope.testrunner.layer.UnitTests in 0.000 seconds.
  Running:
.Question: Enter version [0.1]: 
Our reply: <ENTER>
Checking data dict
Question: OK to commit this (Y/n)? 
Our reply: <ENTER>
Changelog of tha.example
=====================

0.1 (2015-05-07)
----------------
from setuptools import setup, find_packages
import os.path

version = '0.1'



Failure in test /Users/mauritsvanrees/buildout/tools/src/zest.releaser/zest/releaser/tests/prerelease.txt
Failed doctest test for prerelease.txt
  File "/Users/mauritsvanrees/buildout/tools/src/zest.releaser/zest/releaser/tests/prerelease.txt", line 0

----------------------------------------------------------------------
File "/Users/mauritsvanrees/buildout/tools/src/zest.releaser/zest/releaser/tests/prerelease.txt", line 45, in prerelease.txt
Failed example:
    prerelease.main()
    - Question: Enter version [0.1]:
    - Our reply: <ENTER>
    - Checking data dict
    - Question: OK to commit this (Y/n)?
    - Our reply: <ENTER>
----------------------------------------------------------------------
File "/Users/mauritsvanrees/buildout/tools/src/zest.releaser/zest/releaser/tests/prerelease.txt", line 54, in prerelease.txt
Failed example:
    svnhead('CHANGES.txt')
    - Changelog of tha.example
    - =====================
    - <BLANKLINE>
    - 0.1 (...-...-...)
    - ----------------
----------------------------------------------------------------------
File "/Users/mauritsvanrees/buildout/tools/src/zest.releaser/zest/releaser/tests/prerelease.txt", line 60, in prerelease.txt
Failed example:
    svnhead('setup.py')
    - from setuptools import setup, find_packages
    - import os.path
    - <BLANKLINE>
    - version = '0.1'


  Ran 1 tests with 1 failures, 0 errors, 0 skipped in 7.908 seconds.
Tearing down left over layers:
  Tear down zope.testrunner.layer.UnitTests in 0.000 seconds.

Tests with failures:
   /Users/mauritsvanrees/buildout/tools/src/zest.releaser/zest/releaser/tests/prerelease.txt
```

I guess `colorama` really needs to be initialized very early on.

setup.py
mauritsvanrees: Note that `wheel` is currently only a `recommended` dependency. We could promote it a real dependency. But I guess that is not needed, for example when you use zest.releaser to release a non-python project.

mauritsvanrees:
@reinout Tests are fixed. The `init` call could not be moved. I have added coloring for `logger.warn` and `logger.error` too. Do you want to have another look?

reinout:
Looks good. With all these fixes, I *do* want to test it for a few days locally to see if anything crops up.

On the other hand... If releasing zest.releaser works, many things will also work, right :-)

mauritsvanrees:
Agreed on both: testing it a bit more in practice is good and releasing it is good... We are sitting on some good improvements. Maybe indeed give it a few more days. Or at least one.

For a 4.0 release it would be nice if some kind soul, maybe you or me, added support for uploading with twine, issue #59. It kinda fits in with the other improvements. We should not hold our breath though.

reinout:
Twine! That was the name of the upload-thingy that would probably be another core dependency! :-)

Issue #49, opened 2013-11-18
Add recommended extra requirements in setup.py


mauritsvanrees:
We could add a list of recommended extras in setup.py, to be installed with `zest.releaser[recommended]`. This would indicate that we recommend these extra packages if you want to properly release to PyPI. We will want to keep an eye on these, possibly run their tests to see if they break when we change something, and add minimum version requirements if a new version is really needed. I would add these:

```
    zest.pocompile
    check-manifest
    pyroma
```

Possibly `gocept.zestreleaser.customupload`, because I happily use it, but that has nothing to do with releasing to PyPI.

@reinout, what do you think of this?

Cc @iElectric as he wants something like this.

reinout:
Ok, I pointed at this from a couple of issues. How big do we what this list of optional dependencies to be? I originally started zest.releaser with the goal of it being dependency-free and python-only. That's why it is calling git/svn/hg/bzr on the commandline instead of using a python API where available.

So we should be careful, even with a recommendation list. And really really careful with mandatory dependencies (such as twine (#59)) might turn out to be.

I'm not saying an extra dependency here or there is a bad idea, but I do want to be careful. At the same time, this might be a good way to promote great work done by others.

zest.pocompile: I haven't checked what it does, but iirc plone and django expect different things. Plone doesn't want compiled .mo files and django does. Or is it the same now? Long time no plone for me :-)

mauritsvanrees:
zest.pocompile should be fine with both django and Plone. It searches for all LC_MESSAGES directories, no matter where they are (locales, locale, something-completely-different).  In combination with zest.releaser, the idea is explained here: https://github.com/zestsoftware/zest.pocompile#zestreleaser-entry-point
It adds the .mo files to the fresh tag checkout, so they are included in the source dist or wheel, as long as your MANIFEST.in is configured to somehow include those `.mo` files.

Main goal: `.mo` files should *not* be in version control, but *should* be in the final source dist or wheel.  This way, users of the package need to worry about compiling those files themselves.

Note that for example Plone is usually configured to compile the .mo files on startup, but you might run into permission problems if you choose to run the buildout as user Jane and let Plone start with effective user Joe: on behalf of Joe Plone would try to compile the .mo files into a directory where Joe might not have write permission. Plone might not be able to start then, which is why you can configure Plone not to compile .mo files.  If all packages have the .mo files already included, as zest.pocompile does, this is no problem.

mauritsvanrees:
[...]

PR #91, opened 2015-05-05
recommended extra, no Python 2.6, version 4.0

mauritsvanrees:
Combination of three items, that kind-of fit together in my mind, but that may be just me.

- A `recommended` extra, with `wheel`, `check-manifest`, and `pyroma`. Issue #49. As said there: could also simply be a list in the documentation. But when using the `recommended` extra, users automatically get the functionality of these packages, and possibly more in the future, instead of having to somehow discover them.

- Drop Python 2.6 from Travis and tox. See comments in pull request #82, which also serve as a note for all three items in the current pull request.

- Major version number update, due to the above and lots of other recent changes.


mauritsvanrees:
Tests fail because the recommended packages are now used in the tests, which means more output and more questions asked. I'll need to look at that.

reinout:
Looks good! I'll leave merging (together with the other two remaining pull requests) to you for the moment.

Give me a ping if I still need to look at something.

Issue #20, opened 2012-08-14
Abort on PyPi HTTP issues


miohtama:
If PyPi upload fails abort fullrelease

(otherwise reupload attemps become painful when you have done postrelease step)

    Register and upload to pypi (y/N)? yes
    INFO: Running: /Users/mikko/code/plone-venv/bin/python setup.py register -r pypi sdist --formats=zip upload -r pypi
    Showing first few lines...
    running register
    running egg_info
    writing paster_plugins to mfabrik.webandmobile.egg-info/paster_plugins.txt
    writing requirements to mfabrik.webandmobile.egg-info/requires.txt
    writing mfabrik.webandmobile.egg-info/PKG-INFO
    ...
    Showing last few lines...
    unrecognized .svn/entries format in
    Upload failed (503): Server Maintenance

... oops

mauritsvanrees:
If I call `print system(utils.setup_py('My brain hurts'))` currently no error is printed. This is because for python commands we do not show the stderr. See https://github.com/zestsoftware/zest.releaser/blob/3.56/zest/releaser/utils.py#L436
Reason is that this may contain warnings or other fluff that is harmful when we try to extract the version number.
I should probably change a few system calls to quit when there is an error, or to ask the user if she wants to continue.

mauritsvanrees:
I think pull request #90 does this nicely.

mauritsvanrees:
That pull request is merged to master. Will end up in 4.0 soon.
We are not exactly aborting when upload fails, but we print it in red and ask the user if she wants to continue.

mgedmin:
Do you give the user an option to retry?

mauritsvanrees:
No. Would be interesting. I am playing with it now. :-)

mgedmin:
Heh.  I just recently ran fullrelease on a package I didn't have PyPI rights on.  It worked out okay: when the upload failed and zest.releaser asked if I wanted to bump the version to n+1.dev0, I asked for PyPI rights on IRC instead, dug out the temp directory name from xterm scrollback, twine uploaded the sdist, and then let fullrelease continue.

PR #92, opened 2015-05-07
Allow retrying some commands

mauritsvanrees:
When upload to PyPI fails, you may want to try again:

```
running upload
Submitting dist/Products.Poi-2.2.8.tar.gz to https://testpypi.python.org/pypi
warning: no previously-included files matching '*.pyc' found anywhere in distribution
Upload failed (401): Incorrect password
error: Upload failed (401): Incorrect password
There were errors or warnings. Are you sure you want to continue? [Yes/No/Retry]
```

It could be applied to more commands, but the spots where we upload or register or wheel or twine or whatever to PyPI are the obvious ones to add this. This pull request does that.

Thanks to @mgedmin for the idea in issue #20.


zest/releaser/tests/bzr.txt
```diff
@@ -58,9 +58,11 @@ Commit it:
     >>> cmd
     'bzr commit -v -m "small tweak"'
     >>> print(system(cmd))
-    Committing to: TESTTEMP/tha.example-bzr/
-    modified setup.py
-    Committed revision 2.
+    RED Committing to: TESTTEMP/tha.example-bzr/
```
reinout: Using a red color here makes it seem like something went wrong. Wouldn't lightblue or yellow or so be more appropriate?

zest/releaser/utils.py
```diff
@@ -494,6 +518,47 @@ def system(command, input=''):
     return result
 
 
+def retry_command(command):
```
reinout: I'd guess a different name would be better. `retry_command()` sounds like you're re-trying a previously issued command. Ideally, something like `system(cmd, allow_retry=True)`would be the clearest way.

Perhaps renaming `system()` would help? `execute_command()` and `execute_command_with_retry()`? Though I think `allow_retry` is clearest as an option.

zest/releaser/utils.py
```diff
@@ -494,6 +518,47 @@ def system(command, input=''):
...
+        raise RuntimeError(msg)
+    while True:
+        input = get_input(question)
+        if input:
+            input = input.lower()
+            if input == 'y':
+                # Accept the error, continue with the program.
+                return result
+            if input == 'n':
+                sys.exit(1)
```
reinout: Where are we using this? pypi upload, I presume?

The old fullrelease behaviour was to continue processing, including doing prerelease. And to point at the /tmp/ dir with the fresh tag checkout.

Now there's a `sys.exit()`, which means we need to tell people what to do next.

What about doing it automatically? Something like a pypi upload could be handled with an "ouch, error detected. Retrying one time in 10 seconds" and reverting back to the old behaviour otherwise.

reinout: Reason: I'm getting slightly worried about the amount of y/n questions. I think we gained one or two in the last half year already and "just hitting enter" is getting to be quite a habit of everybody, I think :-)

zest/releaser/tests/bzr.txt
mauritsvanrees: Yes, it would be. But this is simply the `bzr` command that prints this to `stderr` where I think `stdout` would have been more appropriate... There are more commands that have chosen differently, so we do gain a few false positives here.

A few lines standing out may not be so bad. Alternatively, we would need to refactor our `system` function to actually read return codes to determine if something went wrong, instead of checking `stderr`.

zest/releaser/utils.py
mauritsvanrees: Yeah, the name is not optimal.

The way I would do an `allow_retry` option, would be to split the `system` command into one function that handles the retry bit, and a second function that focuses on just running the command once, otherwise the amount of logic in the `system` function would get a bit too much for my comfort. But then you have the same situation that there is now in this pull request, except maybe different names.

`system` is not very informative, so `execute_command` may indeed be nicer.

So I give two proposals:

1. Keep current code, rename `system` to `execute_command`, rename `retry_command` to `execute_command_with_retry`.

2. Add `allow_retry` option to `system`. I would say default False, because we only need it in a few cases. To prevent one function from getting too big and difficult, rename current `system` to `_execute_command_once`. Rename `retry_command` to `system` (or `execute_command` for clarity) and give that function the extra option.

So for number 2:

```
def _execute_command(command):
    do what current system(command) does

def execute_command(command, allow_retry=False):
    if not allow_retry:
        return _execute_command(command)
    do what current retry_command(command) does
```

zest/releaser/utils.py
mauritsvanrees: We are indeed using this only in PyPI upload.

I'm not sure there is anything useful we can tell people here if they choose not to continue, except a too obvious 'please investigate'.

We could retry automatically and that could sometimes help. So we could put that in. But the problem could also be wrong PyPI credentials, in which case a retry will not help. I now release some packages as user 'maurits' and some as user 'plone'. If the command fails, I can now edit my `~/.pypirc` and retry and it will actually work.

With this question there is no default answer. So if you just hit enter, you will simply see the same question again.

Would it help to change the possible answers? For example: Retry / Continue / Quit?

zest/releaser/utils.py
reinout: Number two is fine!

zest/releaser/utils.py
reinout: Well, we can stick to yes/no/retry. But then we need to print a little bit of extra info on the result of the choice. Something like this, perhaps less verbose:

Continue ('y'es): we have placed a fresh tag checkout in /tmp/dfdfsdfsdf, you can re-try uploading from there if needed.

Retry: do this if it looks like a temporary internet or pypi outage.

Quit ('n'o): stop completely. Note that the postrelease step hasn't been run then, you need to do that manually.


Something like that?

zest/releaser/utils.py
mauritsvanrees: Number two done. Change pushed.

zest/releaser/utils.py
mauritsvanrees: I am now printing this:

```
    There were errors or warnings.
    You have these options for continuing (first character is enough):
    Yes:   continue with the rest of the program.
    No:    stop completely. Note that the postrelease step has not
           been run yet, you need to do that manually.
    Retry: do this if it looks like a temporary Internet or PyPI outage.
           You can also first edit $HOME/.pypirc and then retry in
           case of a credentials problem.
    Are you sure you want to continue? [Yes/No/Retry]
```

In the function we do not have the information where the tag checkout is. The place where it gets called should have that. Let me try something with raising an exception.

mgedmin: The UX that would make most sense to me is:

> Uploading to PyPI failed.  Retry? [Y/n/q/?]

where the choices are:

Yes -- retry the upload (default, so I can just hit Enter)

No -- don't retry, but continue with the process (which is safe, since zest.releaser will ask me about doing stuff like git pushes so if I wanted to abort, I can still hit ^C then) -- presumably I did a twine upload in a different terminal or something; this is the previous behaviour.

Quit - abort everything.  I'm not sure we even need this, ^C would work.

? would explain the single-letter choices.

Also, a more useful abort (one that would justify a separate option here) would be one that undid the previously-made commits with the version number etc and removed any tags that were created (in case tags get created prior to the upload; I don't remember).  It's something that sounds like a bit of work that can maybe wait for a separate PR.

mauritsvanrees: I like Marius' suggestion. I have implemented it with minor changes. Pull request updated.

Yes, tags get created prior to upload. Reverting the changes could be useful, but tricky. At least tricky to get right for all supported version control systems. Not something for today.

reinout:
Looks good this way!

PR #93, opened 2015-05-08
Support twine for uploading to PyPI

mauritsvanrees:
See issue #59.

As noted there: when a package is not registered to PyPI yet, twine upload will fail. You must either manually login to PyPI and register the package, or use python setup.py register after all. See pypa/twine#21


CHANGES.rst
```diff
@@ -4,6 +4,12 @@ Changelog for zest.releaser
 4.0 (unreleased)
 ----------------
 
+- Added support for twine_.  If the ``twine`` command is available, it
+  is used for uploading to PyPI.  It is added to the ``recommended``
```
reinout: "It is added by the recommended extra" => "It is installed automatically if you use `zest.releaser[recommended]` extra".

doc/source/entrypoints.rst
```diff
@@ -31,6 +31,12 @@ after_checkout
     directory.  Of course, when the user chooses not to do a checkout,
     this entry point never triggers.
 
+before_upload
```
reinout: Good one!

doc/source/uploading.rst
```diff
@@ -91,6 +91,24 @@ You can use no/false/off/0 or yes/true/on/1 as answers; upper, lower or mixed
 case are all fine.
 
 
+Uploading with twine
+--------------------
+
+Since version 4.0, we prefer twine_ for uploading to the Python
+Package Index, because it is safer.  If the ``twine`` command is
```
reinout: "safer." => "safer (becauses it uses https for uploading)." ?

doc/source/uploading.rst
```diff
@@ -91,6 +91,24 @@ You can use no/false/off/0 or yes/true/on/1 as answers; upper, lower or mixed
...
+available, it is used for uploading to PyPI.  It is added to the
```
reinout: "It is added" => "Twine is automatically installed for you if ..."

zest/releaser/release.py
```diff
@@ -102,15 +102,16 @@ def _make_tag(self):
     def _upload_distributions(self, package):
         # See if creating an sdist (and maybe a wheel) actually works.
         # Also, this makes the sdist (and wheel) available for plugins.
+        # And for twine, who will just upload the created files.
         if self.pypiconfig.create_wheel():
-            logger.info("Making a source distibution and wheel of a fresh "
+            logger.info("Making a source distribution and wheel of a fresh "
```
reinout: Good catch.

zest/releaser/utils.py
```diff
@@ -279,6 +279,32 @@ def setup_py(rest_of_cmdline):
...
+def has_twine():
+    """Is the twine command available?
+
+    If twine is available, we prefer it for uploading.  We could try
+    to import it, but it might be importable and still not available
+    on the system path.
+
+    So check if the twine command gives an error.
+
+    Note that --version print to stderr, so it fails. --help prints to
```
reinout: First 'print' => 'prints'

reinout:
Couple of comments, looks good otherwise. OK to merge.

CHANGES.rst
mauritsvanrees: Well, the package will be installed then, but the command script might not be installed, or may not be on your path if zest.releaser is not on the path. With `pip` it is fine, but with `zc.buildout` you need to explicitly add `twine` to the eggs, as we do in the `buildout.cfg` in this pull request:

```
[script]
recipe = zc.recipe.egg
interpreter = python
eggs =
    zest.releaser
    zest.releaser[recommended]
    twine
```

Otherwise the twine package is installed, but the twine command not. Do you know a different incantation of this or another recipe that does this better?

I can change the wording a bit.

reinout:
About your 'zc.recipe.egg' remark: add `dependent-scripts = true` to your zc.recipe.egg part and the commands of your dependencies will also be installed automatically. Works like a charm.

mauritsvanrees:
Ah, right. I have seen it before and maybe used it. Thanks for the reminder. (I will commit that on the retry-command pull request where I have merged master, to avoid more merges of merges.)

reinout:
Why does the monty python sketch about spam spring to minde now? "Merge, merge, merge, merge"... :-)

mauritsvanrees:
We've got a bowl of "merge", or "pull, merge, push, merge, merge", oh and the lovely "fetch, merge, pull, rebase, merge, merge, merge, push, commit, merge."

Don't you have anything without merge in it?

Well, we have "pull, merge, commit, commit, commit, push," which does not have that much merge in it.

...

PR #97, opened 2015-06-01
Full Python 3 support and improved support for unicode messages and filesystems.

mitchellrj:
All tests passing locally on Python 2.7 & Python 3.4. Travis updated to run both.

If merged and released, a note thanking my employer, Isotoma Limited would be much appreciated.


buildout.cfg
```diff
@@ -37,7 +37,7 @@ zope.testrunner = 4.4.3
 [script]
 recipe = zc.recipe.egg
 dependent-scripts = true
-interpreter = python
```
reinout: Is this really needed for 3.4 support? Looks like an unrelated change. I'm personally used to having bin/python in all my buildouts. Is there a good reason to do it otherwise?s

CHANGES.rst
```diff
@@ -1,11 +1,19 @@
 Changelog for zest.releaser
 ===========================
 
-4.1 (unreleased)
+5.0 (unreleased)
```
reinout: Yeah, python 3 support warrants a big version bump :-)

buildout.cfg
mitchellrj: Nope you're right, this is an unrelated change. I added it because I have a habit of using virtualenvs even when using buildout and in this case, buildout tries to overwrite the python interpreter created by virtualenv which breaks things. Happy to change it back though and keep it as a local modification.

zest/releaser/bzr.py
```diff
@@ -24,10 +24,10 @@ def name(self):
         return dir_name
 
     def available_tags(self):
-        tag_info = execute_command('bzr tags')
-        tags = [line[:line.find(' ')] for line in tag_info.split('\n')]
+        tag_info = execute_command([u'bzr', u'tags'])
+        tags = [line[:line.find(u' ')] for line in tag_info.split(u'\n')]
         tags = [tag for tag in tags if tag]
-        logger.debug("Available tags: %r", tags)
+        logger.debug(u"Available tags: {0!r}".format(tags))
```
reinout: You don't need to do .format() in log messages.

mitchellrj: You'll see a lot of this in this pull request. Mostly added to make consistent use of PEP-3101 rather than a mixture of that and string interpolation, however this is just a style preference and again I'm happy to change it back for logging.

zest/releaser/utils.py
```diff
@@ -138,40 +200,40 @@ def ask(question, default=True, exact=False):
...
             raise RuntimeError(msg)
-        logger.debug("Auto-responding '%s' to the question below.",
-                     default and "yes" or "no")
+        logger.debug(u"Auto-responding '{0}' to the question below.".format(
+            default and u"yes" or u"no"))
         logger.debug(question)
         return default
-    while True:
-        yn = 'y/n'
+    while 1:
```
reinout: The original "while True" is also good, right? At least, that's the version I see most often.

mitchellrj: This is force of habit for me ``while 1`` is marginally quicker since the interpreter does not have to resolve the value of the ``True`` symbol.

reinout:
Ok... wow! What an incredible amount of work! Thanks a lot!

There were a couple of small things that I noticed that I commented on already, but the pull request is way to big for me to review right now. I'm at the djangocon.eu and have to leave in a few minutes for dinner :-)

I can look at it during the sprints on thursday and friday.

Can you add yourself to the CREDITS.rst file? Feel free to include a link to the company there :-)

Three questions right away from someone that hasn't done a lot of python 3 porting:

- Wouldn't "from future import unicode_strings" have been handier than the "u" prefix everywhere? 

- Are the shlex and AST changes related to the python 3 porting? I can imagine they made it easier.

- Is it really python 3.4 only? No 3.3? I don't know much about the adoption rates of those. 3.4 might be perfectly fine.

I'm going to learn a lot from reviewing this pull request!

@mgedmin, you've done python 3 porting more often. Can you take a quick cursory look to see if it looks OK?

zest/releaser/bzr.py
reinout: I don't know what is standard in 3.4. I'm still used to %s :-) 

Could you provide a link or so? I might have to change my preference.

buildout.cfg
reinout: Oh, in that case bin/py is fine. I just tab-complete it anyway.

mitchellrj:
Thanks for your quick response to this request!

> Wouldn't "from future import unicode_strings" have been handier than the "u" prefix everywhere?

I find the "u" prefix is more explicit and should help future contributors who are more familiar with Python 2.x realise immediately what kind of string something is meant to be.

> Are the shlex and AST changes related to the python 3 porting? I can imagine they made it easier.

The shlex changes are unrelated but made it easier (and more secure!). The AST stuff again isn't directly related but made things easier.

> Is it really python 3.4 only? No 3.3? I don't know much about the adoption rates of those. 3.4 might be perfectly fine.

It should also support Python 3.3 without issue, though I've not tested it on Python 3.3. It definitely won't support Python 3.0 - 3.2 inclusive. I'll add 3.3 to the Travis configuration now and see how it goes!

zest/releaser/bzr.py
mitchellrj: [PEP 3101](https://www.python.org/dev/peps/pep-3101/) & [str.format in the Python 3 docs](https://docs.python.org/3/library/stdtypes.html#str.format).

The [object.\__format\__ special method](https://docs.python.org/3/reference/datamodel.html#object.__format__) is also particularly fun.

zest/releaser/bzr.py
```diff
@@ -40,24 +40,25 @@ def tag_url(self, version):
...
-        return "bzr diff -r tag:%s..-1" % version
+        return [u"bzr", u"diff", u"-r", u"tag:{0}..-1".format(version)]
 
     def cmd_log_since_tag(self, version):
-        return "bzr log -r tag:%s..-1" % version
+        return [u"bzr", u"log", u"-r", u"tag:{0}..-1".format(version)]
 
     def cmd_create_tag(self, version):
-        return 'bzr tag %s' % version
+        return [[u'bzr', u'tag', version]]
```
mgedmin: Why the double nesting?  Other commands don't do that.

zest/releaser/git.py
```diff
@@ -32,12 +33,14 @@ def name(self):
         # No setup.py? With git we can probably only fall back to the directory
         # name as there's no svn-url with a usable name in it.
         dir_name = os.path.basename(os.getcwd())
+        if not isinstance(dir_name, six.text_type):
+            dir_name = dir_name.decode(sys.getfilesystemencoding())
```
mgedmin: Maybe use `.decode(..., 'surrogateescape')` here [like Python 3 does](https://docs.python.org/3/library/os.html#file-names-command-line-arguments-and-environment-variables)?  (I expect it'll blow up later anyway, when things that expect valid Unicode strings encounter unpaired surrogates, but at least it'll blow up the same way on Python 2 and 3.)

zest/releaser/git.py
```diff
@@ -61,85 +64,92 @@ def tag_url(self, version):
...
             # to work.
-            logger.warn("We haven't been chdir'ed to %s", checkout_dir)
+            logger.warn(u"We haven't been chdir'ed to {0}".format(checkout_dir))
             sys.exit(1)
-        return 'git checkout %s && git submodule update --init --recursive' % \
-            version
+        return [
+            [u'git', u'checkout', version],
+            [u'git', u'submodule', u'update', u'--init', u'--recursive']
+            ]
```
mgedmin: I believe pep8-the-tool complains unless you align the closing `]` with `return`.  I also prefer that alignment, and also a trailing comma after the last item in a multi-line list, if the closing bracket is on a different line.

zest/releaser/hg.py
```diff
@@ -21,14 +24,17 @@ def name(self):
         # No setup.py? With hg we can probably only fall back to the directory
         # name as there's no svn-url with a usable name in it.
         dir_name = os.path.basename(os.getcwd())
+        if not isinstance(dir_name, six.text_type):
+            dir_name = dir_name.decode(sys.getfilesystemencoding())
```
mgedmin: I've seen this code before.  Refactor: extract a function, move it into utils?

zest/releaser/hg.py
```diff
@@ -21,14 +24,17 @@ def name(self):
...
     def available_tags(self):
-        tag_info = execute_command('hg tags')
-        tags = [line[:line.find(' ')] for line in tag_info.split('\n')]
+        tag_info = execute_command([u'hg', u'tags'])
+        tags = [line[:line.find(u' ')] for line in tag_info.split(u'\n')]
         tags = [tag for tag in tags if tag]
-        tags.remove('tip')  # Not functional for us
-        logger.debug("Available tags: %r", tags)
+        while u'tip' in tags:
+            tags.remove(u'tip')  # Not functional for us
```
mgedmin: Now I'm curious about the `while`.

(Suggest `tags = [tag for tag in tags if tag and tag != u'tip']` instead.)

zest/releaser/longtest.py
```diff
@@ -1,40 +1,49 @@
...
-        error = os.system('rst2 html %s > %s' % (filename1, filename2))
-    if error:
+    longdesc = _execute_command(utils.setup_py([u'--long-description']))
+    try:
+        rst = publish_string(
+            source=longdesc,
+            writer_name='html',
+            enable_exit_status=False)
+    except SystemExit:
+        # Stupid docutils throws SystemExit
```
mgedmin: For extra fidelity you could use [readme](https://pypi.python.org/pypi/readme), which is currently used by PyPI (as well as its successor Warehouse) to render long_description strings.

I've encountered long_descriptions that would render fine with plain docutils but fail on PyPI.

doc/source/developing.rst
```diff
@@ -16,9 +16,8 @@ zest.releaser checkout, do this::
 Python versions
 ---------------
 
-The tests currently pass on python 2.6 and 2.7. Using
-2.6.2 specifically will give test failures. Travis continuous integration
-tests 2.6 and 2.7 for us automatically.
+The tests currently pass on python 2.7 and 3.4. Travis continuous
+integration tests 2.7 and 3.4 for us automatically.
```
mgedmin: Does this drop 2.6 compat?  (I'm not complaining, just wondering why that isn't mentioned in the changelog.)

zest/releaser/git.py
mitchellrj: Good idea.

zest/releaser/bzr.py
mitchellrj: Another VCS needs to return multiple commands from this method. It's better
to have a consistent signature than returning mixed results depending on
which VCS one is using.

zest/releaser/hg.py
mitchellrj: Sure thing.

zest/releaser/git.py
mitchellrj: No problem.

zest/releaser/hg.py
mitchellrj: That would be quicker, thanks!

mgedmin:
I can't (force myself to) review all this.  All the `u''` changes are Too Much and, I'd say, unnecessary.

Now I'm not saying mine is the only valid strategy, but I always had good results by
- preferring native strings for pure ASCII literals
- adding the odd `u` prefix here and there for things that must be unicode (e.g. because zope.schema insists)
- using `from __future__ import unicode_literals` as a last resort, when circumstances force me to make most of the literals Unicode (this is a bit of a bug magnet; e.g. some WSGI servers don't like to find three Unicode strings in the WSGI env that's otherwise full of bytestrings; also figuring out how to use future imports in doctests is fun!)

zest/releaser/longtest.py
mitchellrj: TIL thanks!

doc/source/developing.rst
mitchellrj: Not explicitly, just updating the docs to reflect statements made elsewhere
on the release of 14.0.

mitchellrj:
I have tried to use unicode strings explicitly for text-ish content and
those strings which may be unicode in some circumstances (e.g. filesystem,
contents of user-controlled files). Others have been left untouched. If you
spot any which would never be unicode and aren't text-like, that's a
mistake on my part.

zest/releaser/utils.py
mgedmin: I prefer `while True:`.  The speed difference is 20 *microseconds* on Python 2 and disappears completely on Python 3, which forbids assignments to `True` and `False`.

zest/releaser/bzr.py
mgedmin: I'm worried about the inconsistency between different methods of the same class.  Why not make all of them use the same level of nesting?  Or, why not check the types dynamically and allow both `[cmd]` and `[[cmd1], [cmd2]]` from each method?

zest/releaser/utils.py
```diff
@@ -31,6 +49,42 @@ def loglevel():
     return logging.INFO
 
 
+def read_text_file(filename, encoding=None):
+    # Unless specified manually, We have no way of knowing what text
+    # encoding this file may be in.
+    # The 'open' method uses the default system encoding to read text
+    # files in Python 3 or falls back to utf-8.
```
mgedmin: It's a shame that [tokenize.open](https://docs.python.org/3.4/library/tokenize.html#tokenize.open) and [tokenize.detect_encoding](https://docs.python.org/3.4/library/tokenize.html#tokenize.detect_encoding) never made it into Python 2.7.

zest/releaser/bzr.py
mitchellrj: I did consider this but wasn't sure if changing the signature for all of
them would be too far.

zest/releaser/utils.py
mitchellrj: Indeed. IIRC parts of chardet did so on second thoughts, I should really
use those where available.

zest/releaser/vcs.py
```diff
@@ -204,33 +246,41 @@ def _update_version(self, version):
...
+                lines[position[0]]))
+            lines[position[0]] = (
+                lines[position[0]][:position[1]] +
+                repr(version) +
+                lines[position[0]][position[2]:]
+                )
+            contents = u'\n'.join(lines)
+            open('setup.py', 'w').write(contents)
+            logger.info("Set setup.py's version to {0}".format(version))
+            return
```
mgedmin: I don't see where you're handling
```python
setup(
    ...
    version='1.2.3.dev0',
    ...
)
```

which is not an assignment.

The old code used to handle that (that's what the `# oh, probably '    version = 1.0,' line.` comment was about).

mitchellrj: In the find_assignment method it also looks for "keyword"  which behaves
similarly.

mgedmin: Ah!  Maybe a better name would be less confusing (but I can't offer one offhand).

How is the trailing comma preserved?  `position[2]:`?

mitchellrj: The comma should be at ``position[2]``. That said, this does rely on ``repr(value)`` being consistent and it may not be (``u'1.0'`` vs ``'1.0`` in Python 3; ``'1.0'`` vs ``b'1.0'`` in Python 2 - it could also use ``r'1.0'`` and other combinations, too which ``repr`` would render differently). I need to work in a regexp here as well to combat this, but using the AST to find the starting position is still a worthwhile change.

Thanks for prompting me to rethink this.

mgedmin: Would it be possible to modify the AST and write out the file,instead of attempting textual substitutions?  lib2to3 works that way, IIRC.

mitchellrj: Finding a consistent way to write out the AST is troublesome and only supported by third-party packages. I've ended up with an implementation using tokenizer which should be better. Incoming shortly...

reinout: Mitchell: don't put too much effort into it yet. I get the feeling that you're trying to make this piece of code very efficient.

The end result is code that is, on first sight, hardly readable for me. "position", "col_offset"... To me, it looks like it is hard to maintain that way. A good old simple `lines[...] = good_version` is much more readable.

Perhaps AST parsing is fine, but a core principle of zest.releaser, to me, has always been to keep it as simple as possible. That way, exceptions are easier to handle. 

So: I don't mind a bit of brute-force iterating and a bit of regex'ing. It doesn't need to be performant, it just needs to work. And I especially should be able to read it in 2018 to fix some weird bug. So readability counts much more than a bit of purity or speed.

Just wanted to say that before you pour in even more changes into this super-big pull request :-)

mgedmin: (I get the impression that the goal here is reliability rather than performance, but yeah, maintainability might be more useful in practice.)

mitchellrj:
Nope, the number of values is the same.

mitchellrj:
(Rebased on master)

mitchellrj:
Build failure above is only down to PyPI connectivity problems - it needs a repository owner to hit rebuild.

mitchellrj:
Anything more that's directly relevant to this pull?

image

;)

zest/releaser/vcs.py
reinout: And in a few years time, setup.py will probably be a thing of the past, replaced by setup.cfg. There's no AST in there :-)

mitchellrj: Gotcha, that makes sense. I'll take a look at simplifying it or reverting
to the regex method and keeping this as a separate pull request.

zest/releaser/utils.py
```diff
@@ -433,12 +530,16 @@ def run_entry_points(which_releaser, when, data):
...
-    if command.startswith(sys.executable):
+    if isinstance(command, six.string_types):
+        # BBB
+        command = shlex.split(command)
+
+    logger.debug(u"Running command: {0}".format(cmd_to_text(command)))
+    if command[0] == sys.executable:
         env = dict(os.environ, PYTHONPATH=os.pathsep.join(sys.path))
-        if ' upload ' in command or ' register ' in command:
+        if u'upload' in command or u'register' in command:
```
reinout: You've changed the meaning here. I don't remember exactly why there was a space originally behind the opening quote, but I assume it was to fire the "if" only for the actual setuptools upload and register commands, not for "mupload", which was a command from a setuptools extension.

reinout:
See my comment here: https://github.com/zestsoftware/zest.releaser/pull/97/files#r31710024

This is the kind of small change that was almost hidden from me because the pull request is just too big and too full of unrelated changes. Nice that you added the readme package as a dependency, but that should have been a separate pull request. Now it is mixed in with everything else. Same with the AST parsing. And several other changes you made that weren't strictly necessary just to get python3 running.

I'm not saying that those unrelated changes themselves are bad or good, it is just that it all adds up to being the biggest pull request I've ever seen in any code I've ever touched.

I'm just not smart enough to keep 10 different kinds of changes in my head at the same time when reviewing the patch.

Yes, there are tests, but they've also been changed a lot. Probably mostly `u''` stuff, but still. 

Bottom line: this pull request is simply too big. If I'd have to merge it my only option would be to copy-paste stuff, basically. 

I'm now at the djangocon.eu sprint and showed the pull request to someone else and, yes, it is too big and should be split into several smaller ones. What would be handiest is to take it in steps:

- Fix all the print statements and the .format() changes (might be two separate ones).

- All those `u` strings all over the place? No, that's not really explicit because as there are quite a number of strings without either a `u` or a `b`. Those are strings in 2.7 and unicode in 3.0. Better to just do the "future" import and only mark strings that really should be `b` as such. Much cleaner to look at. See https://docs.djangoproject.com/en/1.8/topics/python3/#unicode-literals . (Personally, I'd like zest.releaser to follow the strategy used by Django). Note that your pull request is very valuable as you've identified all the places where a `b` is necessary! In any case, a separate pull request to add the future import and mark the necessary strings.

- The bare minimum of necessary six imports and other changes to get it working on 3.x

- Then, separate from each other, the rest of the improvements like shlex and readme.

Of course I don't mind doing some/most of the work. In any case we should make sure we're not working on the same things at the same time. How do you propose to do this?

I'm off to lunch in 20 minutes and afterwards there's a workshop I want to attend. I'm probably back at 16:00 UK time or so.

zest/releaser/utils.py
mitchellrj: This has changed because the argument has changed. It is now a list of command parts (which is then escaped properly) rather than a single string containing the full command (which may not be escaped properly).

reinout:
Ok, I'm closing this one as it is mostly already included via the other pull requests, ok?

reinout:
I looked again at the log messages. See for example http://stackoverflow.com/q/13131400/27401

Core: I want to stick with `logger.info("%s went wrong", 42)` 

and not `logger.info("{} went wrong".format(42))` 

and really really not `logger.info("%s went wrong" % 42)`

The last two both have the disadvantage that the formatting is done always, even if the message doesn't even need to be rendered.

For backwards compatibility, it apparently isn't possible to allow `{}` as the default style.

PR #100, opened 2015-06-02
Stricter version regexps

### LINKED ISSUE #99: Gets confused by PEP-426 markers in setup.py
I'm trying to make a ZODB release and this is what zest.releaser just offered me:
```
INFO: Set setup.py's version to '4.2.0b2'
INFO: Changed version from '4.2.0.dev0' to '4.2.0b2'
INFO: History file CHANGES.rst updated.
INFO: The 'git diff':

diff --git a/CHANGES.rst b/CHANGES.rst
index b3d8319..d48c96b 100644
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,7 +2,7 @@
  Change History
 ================
 
-4.2.0b2 (unreleased)
+4.2.0b2 (2015-06-02)
 ====================
 
 - Stop dynamic computation of install_requires in setup.py: this doesn't work
diff --git a/setup.py b/setup.py
index cb81a75..e6564af 100644
--- a/setup.py
+++ b/setup.py
@@ -153,7 +153,7 @@ setup(name="ZODB",
       tests_require = tests_require,
       extras_require = {
         'test': tests_require,
-        ':python_version != "2.6"': 'zodbpickle >= 0.6.0',
+        ':python_version='4.2.0b2',
       },
       install_requires = [
         'persistent >= 4.1.0',


OK to commit this (Y/n)? 
```

That is not the line containing version number.

[The version number is here](https://github.com/zopefoundation/ZODB/blob/0f7d2a83426c82f0e0b9ae63503060476dc38b1e/setup.py) (that's a permalink to the right version of setup.py if you want to reproduce).

mgedmin:
The match should be anchored to the beginning of the line, because the 
replacement will replace the entire line.

The match should not allow non-whitespace characters around the '=', except
for the beginning of the string literal.

Fixes #99.

### DIFF
--- a/zest/releaser/vcs.py
+++ b/zest/releaser/vcs.py
@@ -8,12 +8,17 @@ from zest.releaser import pypi
 from zest.releaser import utils
 
 VERSION_PATTERN = re.compile(r"""
-version\W*=\W*   # 'version =  ' with possible whitespace
+^                # Start of line
+\s*              # Indentation
+version\s*=\s*   # 'version =  ' with possible whitespace
+['"]             # String literal begins
 \d               # Some digit, start of version.
 """, re.VERBOSE)
 
 UNDERSCORED_VERSION_PATTERN = re.compile(r"""
-__version__\W*=\W*   # '__version__ =  ' with possible whitespace
+^                    # Start of line
+__version__\s*=\s*   # '__version__ =  ' with possible whitespace
+['"]                 # String literal begins
 \d                   # Some digit, start of version.
 """, re.VERBOSE)
 


reinout:
Are you OK with using master for now? I'm at a conference and I'll have time for a release at the sprints on thursday/friday. Together with the python 3 pull request.

Otherwise perhaps @mauritsvanrees can make a quick release.

mgedmin:
This doesn't block me in any way -- the workaround is to alt-tab to a different terminal and fix the bad setup.py before letting zest.releaser proceed with the commit.

mauritsvanrees:
For the record: this fix was released in 5.0. Thanks.

PR #101, opened 2015-06-04
Python 3 support alone

mitchellrj:
See #97. Factored out just the Python 3 support parts and implemented separately, using ``unicode_literals`` and with existing feedback taken on board.


reinout:
Looking at it right now!

mitchellrj:
The Python 3.4 tests seem to have failed due to some problem with Travis rather than any problems with the code.

reinout:
Agree, looks like a travis problem. Previous 3.4 runs have run just fine. If it runs OK on jenkins on 3.3, that's good enough proof for me :-)

zest/releaser/baserelease.py
```diff
@@ -1,6 +1,9 @@
 """Provide a base for the three releasers"""
 
+from __future__ import unicode_literals
+
```
reinout: Note: I normally don't put a blank line in here as it are all stdlib imports.

But... you don't need to change it, I'm planning to run an automatic tool over the imports anyway as there are multiple places where they're not properly sorted.

zest/releaser/tests/test_setup.py
```diff
@@ -58,9 +62,16 @@
     # clearly.  This catches Fore.RED and Fore.MAGENTA.
     # Note the extra backslash in front of the left bracket, otherwise
     # you get: "error: unexpected end of regular expression"
-    (re.compile(r'\x1b\[31m'), 'RED '),
-    (re.compile(r'\x1b\[35m'), 'MAGENTA '),
-])
+    (re.compile(re.escape(Fore.RED)), 'RED '),
```
reinout: Neat!

zest/releaser/utils.py
```diff
@@ -88,7 +168,10 @@ def get_next_answer(self):
 def get_input(question):
     if not TESTMODE:
         # Normal operation.
-        return raw_input(question).strip()
+        result = input(question)
```
reinout: This is something that needs fixing. input() on python2 *evaluates* what you put in. See http://stackoverflow.com/a/15129556/27401

If I enter `1.2rc` as a version string, python evaluates it under 2.x and gives a syntax error.

mitchellrj: Note the import above from six - that should avoid this case.

reinout: `from six.moves import input` ought to solve it.

reinout: Ah? Ok, I missed that.

reinout:
Looks fine. Thanks for re-working the pull request. This one was easy to review!

mitchellrj:
More to follow with the other changes! Thanks.

reinout:
I loved the readme change, but for instance the one with the AST and the  `position[3][1]` was not very readable (and not really worth all the trouble, imho). 

Oh, and the `.format()` was fine, too.

The change that modified the commandline strings into a list of items was one of the things I also didn't really like on first sight. It sounded like something that's theoretically better, but as we're not having any problem with it right now... Readability counts.

With the commandline change, I could imagine that we might need/want to change the way we run commandline stuff now that we only have 2.7/3.3+ to worry about. On the other hand, the current approach works flawlessly...

So: for the ones where I made a comment, it might be better to first open an issue to explain why it is necessary and/or very useful so that @mauritsvanrees can also have a look. Ok?

The rest is very welcome!

reinout:
Just a note: all the tests run fine locally (python2.7 OSX and python 3.4 ubuntu).

reinout:
Hurray, http://reinout.vanrees.org/weblog/2015/06/05/zest-releaser-python-3.html

I've gone ahead and released 5.0 :-)

zest/releaser/tests/test_setup.py
mauritsvanrees: Ah, I wanted to do something like this when I added color support, but did not know how. Thanks!

mauritsvanrees:
Thanks a lot for this, Richard! Sorry, I did not take the time until now to look at it, but it looks good. And the tests pass.

A question about wheels, if anyone knows.  We are currently releasing zest.releaser itself as old fashioned source dist and as a shiny new wheel. Looks like Reinout released 5.0 with a zest.releaser that was itself using Python 3, because when you look at the list of files near the bottom of https://pypi.python.org/pypi/zest.releaser/5.0 you see zest.releaser-5.0-py3-none-any.whl. A recent pip on Python 3 uses this wheel, but on Python 2.7 not. A universal wheel would presumably fix this. Question is: is this safe? I thought I had read somewhere that this should not be done when you use `six` for reaching compatibility between 2 and 3. But I do not see that remark on http://pythonwheels.com. Ah, I am mixing things up: it says you should not use wheels if you are using 2to3. We are not using this.

So: would it be safe to start creating universal wheels for zest.releaser?

In setup.cfg:

```
[bdist_wheel]
universal = 1
```

reinout:
Only one way to find out: why haven't you released a 5.0.1 yet? :-)

mauritsvanrees:
I tried it locally and it worked fine. I compared the resulting wheels and found no difference at all (not counting irrelevant differences in sort order of the various requirements). So we can do that. I have updated the setup.cfg.

You say 5.0.1. Shall we just make that 5.1?

reinout:
5.1 is fine

mauritsvanrees:
I have released 5.1 as sdist and universal wheel:

```
INFO: This package is registered on PyPI.
Register and upload to pypi (Y/n)? 
INFO: Running: twine upload dist/* -r pypi
Uploading distributions to https://pypi.python.org/pypi
Uploading zest.releaser-5.1-py2.py3-none-any.whl
Uploading zest.releaser-5.1.tar.gz
```

PR #102, opened 2015-06-04
Properly escape all command content

mitchellrj:
This breaks all commands down into individual arguments and stops running commands in the shell by default. This ensures that the contents of arguments is properly escaped when run.

Before this change, a peculiar commit message, version string or filename could break the command and cause unexpected results.

This pull request also makes the return values from all ``cmd_*`` method consistent, returning a list of commands, rather than relying on a type test to determine if there is a list of commands or a single command.

See also #97 from which this was factored out.


reinout:
Like I said before, I'm not in favour of this change.

The readability is lowered by this change. And I don't like lists of lists of command-parts.
And I really don't mind a string-or-list check if it is just for a single command (some mercurial one, if I remember correctly).

@mauritsvanrees, what's your opinion?

mitchellrj:
Python 3.3 failed due to a network error, Python 3.4 failed due to #105.

mitchellrj:
@reinout 

I suppose my key reason for this is

> Before this change, a peculiar commit message, version string or filename could break the command's string interpolation and cause unexpected results.

The use of ``'command arg1 "%s"' % 'long arg 2'`` is fairly dangerous and should be avoided, particularly when running commands in shell mode.

reinout:
I think there's no need to fear for unproperly escaped arguments. The simple reason is that we ourselves provide those arguments!

There are some user-provided arguments like the project name and possibly an alternative pypi server and a version number. But... those aren't allowed to contain weird quotes! DNS doesn't allow quotes. And if you've got a quote in your version number, I don't mind that the command line fails.

So: I don't think it is actually a problem.

I well keep the shlex module in mind for other projects, though!

mauritsvanrees:
I agree with Reinout: we do not need this currently. If actual problems surface, we can revisit this.

PR #103, opened 2015-06-04
Use the "readme" library for long description testing.

mitchellrj:
(Factored out from #97)


reinout:
Looks good. I have some comments about the error messages, but I'll just do that in a separate pull request.

PR #104, opened 2015-06-04
Readme fixes

reinout:
Some improvements & fixes for #103 

- Clearer error messages.

- Long description error is printed again (only printing "it is invalid" isn't helpful).

- The tempfile didn't have a `.html` extension (only `html` as a postfix, not an extension)


mitchellrj:
:+1:

PR #105, opened 2015-06-05
Fix Travis configuration for Python3.4

mitchellrj:
I blame my stupid fat fingers from #101.


reinout:
I just noticed the same comma 5 minutes ago and fixed it directly on master already :-)

PR #107, opened 2015-06-05
Use AST & tokenize to parse Python files rather than relying on regular expressions.

mitchellrj:
Gains: more reliable
Losses: less readable

This has been made more readable than it was in #97.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,9 @@ Changelog for zest.releaser
   descriptions when we test and render them.
   [mitchellrj]
 
+- Use the AST and tokenizer when parsing and writing Python code files.
+  [mitchellrj]
+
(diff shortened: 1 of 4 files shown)


reinout:
The current code has survived 8 years of people throwing corner cases at them. So making it even more reliable isn't really something that is needed. The readability has improved compared to the previous version, but, as you say, it *is* less readable than the previous version.

Can you come up with a test that would fail with the current code and work with your version?

mauritsvanrees:
I am going to have to agree with Reinout here too. It has not been needed yet. And if it *is* needed, the best way forward is to fix the silly version setting in the offending package.  If someone comes along with a package that needs this fix and that cannot fix its own version, we can revisit this.

Issue #27, opened 2012-10-09
Add support for Subversion1.7


woutervh:
Hello,

Using the latest zest.releaser 3.39, I got surprised by following message when running "fullrelease" in my svn-trunk-folder:

> CRITICAL: No version control system detected.

However nothing is wrong with my svn-checkout. I'm using a recent  svn:

> svn --version
>    svn, version 1.7.6 (r1370777)

The problem is that zest.releaser searches for a ".svn"-directory in the current location (choose.py)
That assumption no longer holds in subversion1.7.
 From the SVN-book, http://svnbook.red-bean.com/en/1.7/svn-book.pdf  (page 10):

> Prior to version 1.7, Subversion maintained .svn administrative subdirectories in every versioned  directory of your working copy. Subversion 1.7 offers a completely new approach to how working  copy metadata is stored and maintained,  and chief among the visible changes to this approach is  that each working copy now has only one .svn subdirectory  which is an immediate child of the root  of that  working copy."

Maybe it's easier to checkout the output of  "svn info" executed in the current directory.

PR #111, opened 2015-06-10
Add support for projects with setup.py located in subdirectories.

blampe:
Consider a  repo with a file structure like this:
```
repo/
  foo/
      setup.py
  .git
```

It may not be the recommended structure for a python project, but it happens in practice.

Running `fullrelease` from within `repo`:

```
$ fullrelease
INFO: Starting prerelease.
CRITICAL: No version found.
```

Running it from within `foo`:

```
$ fullrelease
INFO: Starting prerelease.
CRITICAL: No version control system detected.
```

With this change, from within `repo`:
```
$ fullrelease --base foo
INFO: Starting prerelease.
...
```

I thought about maybe trying to be smart by recursing to find the file, but I prefer this explicitness. Let me know what you think!

### DIFF
--- a/zest/releaser/utils.py
+++ b/zest/releaser/utils.py
@@ -32,7 +32,7 @@ WRONG_IN_VERSION = ['svn', 'dev', '(']
 MUST_CLOSE_FDS = not sys.platform.startswith('win')
 
 AUTO_RESPONSE = False
-BASE = "."
+BASE = os.curdir
 VERBOSE = False
 INPUT_ENCODING = 'UTF-8'
 if getattr(sys.stdin, 'encoding', None):
--- a/zest/releaser/utils.py
+++ b/zest/releaser/utils.py
@@ -32,6 +32,7 @@ WRONG_IN_VERSION = ['svn', 'dev', '(']
 MUST_CLOSE_FDS = not sys.platform.startswith('win')
 
 AUTO_RESPONSE = False
+BASE = "."
 VERBOSE = False
 INPUT_ENCODING = 'UTF-8'
 if getattr(sys.stdin, 'encoding', None):
@@ -133,7 +134,12 @@ def cleanup_version(version):
 def parse_options():
     global AUTO_RESPONSE
     global VERBOSE
+    global BASE
     parser = ArgumentParser()
+    parser.add_argument(
+        "--base",
+        default=BASE,
+        help="Directory containing setup.py")
     parser.add_argument(
         "--no-input",
         action="store_true",
@@ -150,6 +156,7 @@ def parse_options():
     options = parser.parse_args()
     AUTO_RESPONSE = options.auto_response
     VERBOSE = options.verbose
+    BASE = options.base
 
 
 # Hack for testing, see get_input()
@@ -371,7 +378,9 @@ def setup_py(rest_of_cmdline):
             if unsafe in rest_of_cmdline:
                 executable = 'echo MOCK'
 
-    return '%s setup.py %s' % (executable, rest_of_cmdline)
+    setup_py = os.path.join(BASE, 'setup.py')
+
+    return ' '.join([executable, setup_py, rest_of_cmdline])
 
 
 def twine_command(rest_of_cmdline):
--- a/zest/releaser/utils.py
+++ b/zest/releaser/utils.py
@@ -32,7 +32,7 @@ WRONG_IN_VERSION = ['svn', 'dev', '(']
 MUST_CLOSE_FDS = not sys.platform.startswith('win')
 
 AUTO_RESPONSE = False
-BASE = os.curdir
+PACKAGE_ROOT = os.curdir
 VERBOSE = False
 INPUT_ENCODING = 'UTF-8'
 if getattr(sys.stdin, 'encoding', None):
@@ -134,11 +134,11 @@ def cleanup_version(version):
 def parse_options():
     global AUTO_RESPONSE
     global VERBOSE
-    global BASE
+    global PACKAGE_ROOT
     parser = ArgumentParser()
     parser.add_argument(
-        "--base",
-        default=BASE,
+        "--package-root",
+        default=PACKAGE_ROOT,
         help="Directory containing setup.py")
     parser.add_argument(
         "--no-input",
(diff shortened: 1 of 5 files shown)


zest/releaser/utils.py
```diff
@@ -32,6 +32,7 @@
 MUST_CLOSE_FDS = not sys.platform.startswith('win')
 
 AUTO_RESPONSE = False
+BASE = "."
```
mauritsvanrees: This should be `os.curdir`. Well, it seems to be the same everywhere (https://docs.python.org/2/library/os.html#os.curdir) but seems better to me anyway.

mauritsvanrees:
I guess this would be okay. Tests at least pass. 

Do you have a link to a package where this would be needed?

Does it actually work? I wonder if you would need to change to the BASE directory and then call `python setup.py`.

An update to the docs would be needed to explain this option. One or two lines are probably enough. Add something like that to the changelog too, so people now how to use this new support.

zest/releaser/vcs.py
```diff
@@ -318,7 +320,7 @@ def list_files(self):
         works is handy for the vcs.txt tests.
         """
         files = []
-        for dirpath, dirnames, filenames in os.walk('.'):
+        for dirpath, dirnames, filenames in os.walk(utils.BASE):
```
mauritsvanrees: I think this one should not be done. With this change, zest.releaser will look for a changelog file in utils.BASE, so it will miss the changelog file in the main directory.

blampe:
I tested this on one of our [projects](https://github.com/uber/tchannel) and everything worked as expected, no need to `cd` into the subdirectory to run `setup.py`

> An update to the docs would be needed to explain this option.

I added the option name to the changelog. Any place in particular you'd like to see this in the docs? Maybe `assumptions.rst`?

zest/releaser/vcs.py
blampe: Interesting point! I'd argue that the current behavior is to grab whatever change log is "sitting next to" our `setup.py`, so if the `setup.py` is nested then I'd expect to see a nested change log as well. Change logs higher up in the directory structure might be unrelated to the python package.

Take a "monorepo" like [twitter-text](https://github.com/twitter/twitter-text ), for example. It wouldn't be unreasonable to add a `python` subdirectory and a separate change log for the ruby gem. Searching from the repository root with `--package-root=python` would find that unrelated change log.

That said, I don't feel strongly about this so I got rid of it.

zest/releaser/utils.py
blampe: Fixed and renamed `BASE` to `PACKAGE_ROOT` as I think that's more descriptive.

mauritsvanrees:
Note: if zest.releaser now starts touching the wrong changelog file, you can explicitly point it to a file. Create a setup.cfg file in your package with something like this:

```
    [zest.releaser]
    python-file-with-version = mypackage/some_file.md
```

This is briefly explained in the docs, in versions.rst.  This should probably be in an options.rst file, just like your new option. :-) I can start a file like that on master and copy your changelog entry in it and maybe expand a bit on it when I merge. That should be enough for documentation. I'll handle that.

I still worry about about subtle things going wrong that we have not thought about yet.  We can fix those later as we discover them, so it need not hold back a merge. With your pull request the basics at least work, instead of quitting with a CRITICAL error like above.

But an example I now think about: we try to read a `setup.cfg` in the current directory. See https://github.com/zestsoftware/zest.releaser/blob/5.1/zest/releaser/pypi.py#L46. That should be changed to look in the PACKAGE_ROOT now.

And: in your example project you would basically be releasing tchannel/python but at the same time making a tag of the complete tchannel, which seems a bit weird. I do see that you have tags like py-0.8.3 and node-v2.0.0, so it looks like you are handling that already. I guess I would have opted to make them separate repos, but there may be valid reasons that you do not need to explain. :-)

Out of interest: would you use zest.releaser to also make releases of the node and golang parts? A `setup.py` is only needed for releasing to a python package index. As long as zest.releaser can find a version file and a changelog file it will happily adjust those and create a tag. And possibly you could use our entry points to do extra stuff, like creating some kind of node or golang package and uploading it to wherever those need to be. I am just curious. :-)

@reinout Any thoughts on this pull request?

reinout:
Well, it is a small change and it doesn't inconvenience us really, so in that sense it is OK.

On the other hand, it is a corner case that almost nobody is going to see, I think. And every extra option is an extra option that clutters up the documentation, making it harder to read.

Almost every "let's look in other directories" change has resulted in subsequent bug reports due to unintended consequences.

Would this be an alternative option: if no version control markers (like .git, .bzr, .svn) are found, cannot we look up a few directories? A maximum of two, for instance? That wouldn't need an extra option. (But could have unintended consequences of its own... I guess we should warn like "didn't find a version control system in our current directory, but ../.. looks like we're in a git project".

mauritsvanrees:
Something like that could work too. I've thought about that before: it might be fine to do a release when you are not in the root of your project. Our `choose.py` looks for `.git` etc in the current directory.  For subversion we already have an exception, look at issue #27. So something like that might be okay too.

Specifically for subversion I do wonder if a tag is properly created from the current trunk instead of from `trunk/your/current/directory`, but if this is the case, that error would be in the current code already.

blampe:
Great feedback! Trying to respond to all the points raised.

* `options.rst` sounds good to me. 
* Good catch on `setup.cfg`, made the change to look in `PACKAGE_ROOT`. 
* Yeah, in that project we manually manage our tags, one for each language. It's a little awkward but we make it work. It's nice that releaser provides sane defaults but allows overriding them.
* In theory we could use this to tag our other languages, but in practice I think it would be cumbersome because we work in virtual environments. So the node guys, who are more comfortable with `npm`, would have an extra/unfamiliar step in their release process. On the other hand, once we start tagging more frequently it could gain some more traction. We'll see!
* I agree there may be some corner cases. @reinout's solution is also very reasonable, so I leave it to you folks to decide what you'd like to do :)

mauritsvanrees:
I would be in favor of Reinout's solution if it works. So a bit along the lines of 0d3be488882ef81c9f9b09d0a51a917680c0d9d7.

mauritsvanrees:
I am having a go at that by the way.

blampe:
:+1:

mauritsvanrees:
This is superseded by pull request #117.

PR #115, opened 2015-06-19
Always use --non-interactive for svn commands.

mitchellrj:
If there are certificate errors these are silently dropped into output making for strange behaviour without this option.


mauritsvanrees:
I thought we did this already, but that was with checkoutmanager. Looks good. Thanks. I'll merge.

Issue #109, opened 2015-06-09
Support non-ASCII changelog files


wosc:
Before the Python-3 support came in, zest.releaser seemed to handle non-ASCII changelog files just fine (I never had any issues), but now I get this error:

```
  File "/home/wosc/install/zest/local/lib/python2.7/site-packages/zest.releaser-5.0-py2.7.egg/zest/releaser/prerelease.py", line 71, in execute
    self._write_history()
  File "/home/wosc/install/zest/local/lib/python2.7/site-packages/zest.releaser-5.0-py2.7.egg/zest/releaser/prerelease.py", line 161, in _write_history
    open(history, 'wb').write(contents)
UnicodeEncodeError: 'ascii' codec can't encode character u'\xfc' in position 9914: ordinal not in range(128)
```

I've locally added `contents.encode('utf-8')` as a workaround, but I must confess, I'm not sure what a generic solution would look like -- and what the previous behaviour actually was (I'm guessing some kind of default encoding behaviour?).

reinout:
If I remember correctly, there's a test somewhere that verifies that we can *read* a changelog with non-ascii characters. Shouldn't be too much effort to get my hands on a test that tests the writing back of the same file :-)

I'll see if I can fix it quickly. If it isn't fixed within two days, feel free to post a quick reminder here!

mauritsvanrees:
I had the same issue recently which I fixed by editing the changelog. See https://github.com/plone/Products.TinyMCE/commit/4df527f8b9c7e2f76c57d343afb0a336d7b6f153

With zest.releaser 3.56 it worked fine. With 4.0 it fails with the UnicodeEncodeError.

Using `contents.encode('utf-8')` (in prerelease and postrelease) fixes it for Python 2.7, but then you get `TypeError: must be str, not bytes` on Python 3.

mauritsvanrees:
I think I have fixed it in the above commit for Python 2 and 3.

mauritsvanrees:
BTW, with the above fix, when I do a prerelease of Products.TinyMCE which has a weird and wrong and unnecessary non-ascii `s` in its changelog, I get this as one of the changes:

```
-  Fixeѕ https://dev.plone.org/ticket/10681
+  FixeŃ https://dev.plone.org/ticket/10681
```

and then for a postrelease:

```
-  FixeŃ https://dev.plone.org/ticket/10681
+  FixeĹÂ https://dev.plone.org/ticket/10681
```

[...]

PR #116, opened 2015-06-19
Write file with the same encoding that we used for reading them.

mauritsvanrees:
This is a further fix for issue #109.


mitchellrj:
Looks good to me!

PR #117, opened 2015-06-24
Release from subdir

mauritsvanrees:
This supersedes pull request #111 with a different approach.  I have tried it with dummy projects in git, bzr, mercurial (not subversion actually) and it seems to work fine.

Note that if you want to release a Python project, we still expect that `setup.py` is in the current directory. But this can now indeed be in a sub directory.


blampe:
Awesome, thanks!

PR #120, opened 2015-08-21
small typo-fix svn-command

awello:


reinout:
Thanks for the fix! I've released 5.3, can you check if it works for you now?

PR #122, opened 2015-08-26
enable non-ascii vcs-messages

awello:


zest/releaser/release.py
```diff
@@ -98,7 +98,7 @@ def _make_tag(self):
         for cmd in cmds:
             print(cmd)
             if utils.ask("Run this command"):
-                print(execute_command(cmd))
+                logger.info(execute_command(cmd))
```
reinout: Doing this as a log message breaks the "layout" of the printed information a bit.

Any idea why logger.info() would work fine and print() not? Can you provide some context for us?

awello:
context: I'm playing around with a German-speaking system and subversion - subversions messages are in German and contain "umlauts" (using de_DE.UTF-8) ... with print I'm getting the unicode-encode-error
I'm trying to get the tests running to fix them but I'm not familiar with bazaar (the sanity_check is failing)

mauritsvanrees:
How are you getting zest.releaser to commit a German subversion message?
Are you overriding the commit messages, like `PRERELEASE_COMMIT_MSG` in `release.py`?

I tried it differently, by adding this in my `~/.pypirc`: `[zest.releaser] extra-message=Ich bin ein Ümläüt`. This added the umlauts just fine in a prerelease message.  The `$LANG` on my Mac is set to `nl_NL.UTF-8`, which should not be that different from your German version.

The real Travis failures in the doctests can be hard to detect, but this is the important line that is missing in the output:
https://travis-ci.org/zestsoftware/zest.releaser/jobs/77277096#L546
I think the way things are now, print statements are printed first, and then log messages or stderr. Could be the other way around too. But here for some reason the `Committed revision 6` line is completely gone, although further on the `Committed revision 7` line is printed just fine. So it is a bit weird.

If you want to ignore bazaar for the moment, you can run only the functional subversion tests: `bin/test -t functional-svn.txt`

But for bazaar you probably simply need to do a few simple commands manually once. See
http://zestreleaser.readthedocs.org/en/latest/developing.html

reinout:
@mauritsvanrees : he's not entering german messages himself.

The OS is set to german and the subversion command that we call has non-ascii output like "Wie schön, alles ist wieder ganz neu". *We* then print the output of that command.

So: sounds like something that needs fixing on zest.releaser's side.

reinout:
Apparently `logger.info()` escapes/encodes/whatever the non-ascii output correctly and `print()` does not.

I think we need to look at the output of `execute_command()`. A quick check would be `execute_command("echo ümlàut")`.

mauritsvanrees:
Goes just fine, at least in a pdb:

```
(Pdb) execute_command("echo ümlàut")
u'\xfcml\xe0ut\n'
(Pdb) print(execute_command("echo ümlàut"))
ümlàut

(Pdb) x = execute_command("echo ümlàut")
(Pdb) x
u'\xfcml\xe0ut\n'
(Pdb) x.encode('utf-8')
'\xc3\xbcml\xc3\xa0ut\n'
(Pdb) print(x.encode('utf-8'))
ümlàut
```

Obviously, this assumes utf-8, but @awello is using that.
So: this remains strange for me...

mauritsvanrees:
If this is still a problem, please try with latest version [7.0.0a1](https://pypi.org/project/zest.releaser/7.0.0a1/). Python 3 only, so maybe encoding problems have gone away.

PR #131, opened 2015-09-23
Add PyPy support by making sure to close all the files we write to.

jamadden:
Because PyPy doesn't use a refcounted GC, leaving dangling files open causes the writes to be flushed at some non-deterministic time in the future, meaning that prerelease (and anything that wanted to write to the users files) typically wound up committing an empty file.

Most of the fixes were in the tests but there were several in the actual code itself. I didn't fix all the read() cases in the tests for clarity. Those do leave dangling files and so running out of file-descriptors is a possibility, but not an immediate practical concern.

Note: The svn.txt test fails locally; I think it's just due to a difference in the output by different SVN versions. Let's see what Travis says.


jamadden:
Yay, Travis was all green! So the local SVN failure does appear to be due to differences in SVN versions. For reference, Travis had SVN 1.6.17 (really old) and locally I have version 1.9.1. I know there are documented command-line output changes in [1.8](http://subversion.apache.org/docs/release-notes/1.8.html#output-changes) and [1.7](http://subversion.apache.org/docs/release-notes/1.7.html#output-changes). It looks like the 1.7 changes are the breaking changes here.

mauritsvanrees:
Thanks! I will merge.

Locally we are testing with whatever svn we have on our machines. In my case for svn that is version 1.9.1 on Mac. Tests pass fine there. In `test_setup.py` there is normalization code that should make the tests run with several command versions. Can you paste the failure you got? Unfortunately when there is a doctest failure you will see the entire difference without heeding any test normalization that was done, so you then have to hunt in a too large diff for that one tiny character that makes all the difference...

After merge I will check if it runs un pypy3 as well. And update tox.ini.

jamadden:
Ah, I didn't spot the normalization code. Otherwise, the failures look to be because the newer SVN adds a header line and maybe some whitespace (I agree, debugging doctests like that is hard). Here's what doctest spits out on OS X 10.11:
```
File "//zest/releaser/tests/svn.txt", line 45, in svn.txt
Failed example:
    print(execute_command(cmd))
    - Index: ...
    - --- setup.py      (revision 3)
    - +++ setup.py      (working copy)
    - @@ -36,3 +36,5 @@
    + Index: setup.py
    + ===================================================================
    + --- setup.py	(revision 3)
    + +++ setup.py	(working copy)
    + @@ -36,3 +36,5 @@
    -            'console_scripts': [
    +            'console_scripts': [
    ? ++++++                               ++++++
    -            ]},
    -        )
    - +
    - +a = 2
    +            ]},
    +        )
    + +
    + +a = 2
    + <BLANKLINE>
----------------------------------------------------------------------
File "//zest/releaser/tests/svn.txt", line 110, in svn.txt
Failed example:
    print(execute_command(cmd))
      Index: setup.py
      ===================================================================
    - --- setup.py      (.../tags/0.1)  (revision 7)
    ?             ^^^^^^              ^^
    + --- setup.py	(.../tags/0.1)	(revision 7)
    ? +++++++            ^              ^            ++++++
    - +++ setup.py      (.../trunk)     (revision 7)
    - @@ -38,3 +38,5 @@
    -        )
    + +++ setup.py	(.../trunk)	(revision 7)
    + @@ -38,3 +38,5 @@
    +        )
    +
    +  a = 2
    + +
    + +b = 3
      <BLANKLINE>
    -  a = 2
    - +
    - +b = 3
```

As I'm sure you know, PyPy3 implements Python 3.2 with some extensions (the `u' prefix, for one). It may or may not run. Since 3.2 isn't in the supported list already, and is being end-of-lifed, and the PyPy maintainers are actively working on a 3.3 compatible release (and our organization doesn't use PyPy3 currently), I didn't test it.

mauritsvanrees:
The header line will be caught by the doctest ellipsis (...) so that should be no problem. An extra blank line would be okay. Some extra white space at the end of the line (maybe beginning too) may also be fine, not sure. The real one might be this white space difference inside a line:

```
    - --- setup.py      (revision 3)
    - +++ setup.py      (working copy)
    + --- setup.py  (revision 3)
    + +++ setup.py  (working copy)
```

Anyway, working fine for me.

PyPy3 works fine when using tox. But in Travis we use buildout. The resulting bin/test script fails at startup. Let's not bother with that. I am keeping pypy3 in tox.ini for now, but have removed it from .travis.yml.

PR #135, opened 2015-10-07
Write back changelog with original encoding, just like already happens with setup.py

reinout:


reinout:
@mauritsvanrees, any comments? Otherwise I'll just merge and release.

mauritsvanrees:
Oh, right, I think I saw this but failed to see it was a pull request.
I'll merge and do a release.
That will be the first release from my new laptop, so please check if all is well. :-)

mauritsvanrees:
5.7 released

PR #137, opened 2015-10-21
signal an error on --no-input and nothing changed yet

do3cc:


do3cc:
WIP, I have more changes locally to add.

mauritsvanrees:
When we exit explicitly, we should probably always exit with code 1. We were already doing this in most cases.  I have fixed the two remaining cases as part of pull request #139. It is not really related to the rest of the pull request, except that one case was in the middle of some code that I changed a lot, so I thought I might as well fix the other one (in svn) as well.

Did you have other plans with your pull request?

do3cc:
Yes, I wanted to change the log level to error if --no-input is set.
But I still did not make the tests pass with my very new git...
I still plan to and can create a fresh PR just for this. If you fixed the exitcode elsewhere already, you can close this one.

reinout:
Ok, closing this one.

Note that the automatically-answered questions are now always logged at INFO level, so at least you can see what happens if zest.releaser does a `sys.exit(1)` afterwards.

do3cc:
Just fyi,

we are using zest.releaser code embedded in a script of our own. Maybe there will be PRs pushing sys.exit closer to the method called by the script

mauritsvanrees:
Note that I fixed some possible test failures for git svn yesterday, maybe that helps you too.

PR #138, opened 2015-10-27
When uploading with twine first use the 'twine register' command.

### LINKED ISSUE #128: Use twine register command
When a package is not registered yet, we warn:

```
WARNING: This package is NOT registered on PyPI.
WARNING: Please login and manually register this package on PyPI first.
```

twine 1.6.0 was released yesterday, which includes a register command.  We can call that like this:

```
twine register dist/mauritstestpackage2-1.1.tar.gz 
```

Or for a different PyPI server, something like this:

```
twine register -r testpypi dist/mauritstestpackage2-1.1.tar.gz 
```

So call it pointing to some generated file in the dist directory.

mauritsvanrees:
This needs twine version 1.6.0 or higher, which you get with the `zest.releaser[recommended]` extra.

This fixes #128.

Note that when you are uploading to something else than PyPI, there is no official way of asking the server whether the package is already registered, so we always register in that case. This at least works fine for testpypi.

Potential problem: if someone has not installed `zest.releaser[recommended]` but has installed an earlier version of `twine`, the `register` command will fail.  Possible solutions:
- Make twine a core dependency. We can clean up a bit of code then too.
- Check the version, similar to `utils.has_twine()`


reinout:
Turning twine into a core dependency is fine.

I suspect that https uploading will be mandatory in the next version of pypi ("wheelhouse") anyway.

Can you adjust the PR?

mauritsvanrees:
I made twine a core dependency and cleaned up the other code paths.  I updated our version to 6.0.

CHANGES.rst
```diff
@@ -1,9 +1,19 @@
...
+6.0 (unreleased)
 ----------------
 
+- Made ``twine`` a core dependency.  We now always use it for
+  registering and uploading.  We require at least version 1.6.0.
+  [maurits]
+
+- When uploading with ``twine`` first use the ``twine register``
+  command.  This needs ``twine`` version 1.6.0 or higher, which you
+  get with the ``zest.releaser[recommended]`` extra.
```
reinout: If it is a core dependency, you don't need to install it via "recommended" anymore :-)

setup.py
```diff
@@ -5,7 +5,7 @@
 from setuptools import setup
 
 
-version = '5.8.dev0'
+version = '6.0.dev0'
```
reinout: Good one! Yes, 6.0 it is.

reinout:
One changelog comment. After that you can release it!

mauritsvanrees:
I did some rebasing and updated the docs as they still talked about preferring twine instead of requiring it.
Let's see if Travis does not think I messed something up, and then 6.0 it is.

Issue #142, opened 2015-10-28
twine was not found


jensens:
Looks like it needs twine in path or so? Its in setup.py as dependency, but it is not in buildouts bin directory, so scripts are not installed?

    Register and upload to pypi (y/N)? yes
    INFO: Running: twine register dist/pas.plugins.authomatic-1.0a1.tar.gz
    /bin/sh: 1: twine: not found
    There were errors or warnings.
    Package upload may have failed.
    Retry this command? [Yes/no/quit/?]

After tweaking ``buildout.cfg`` like so:

    [releaser]
    recipe = zc.recipe.egg:scripts
    eggs = 
        zest.releaser[recommended]
        twine

I had the executable in there.

reinout:
Add `dependent-scripts = true` after the `recipe = zc.recipe.egg` line. That installs the scripts of eggs that are dependencies. (So: you don't need to explicitly list `twine`, then).

@mauritsvanrees: hm, this could be a real problem. Can we use twine as a library? Can we list twine's console_script entry point in our setup.py?

reinout:
@jensens, can you retry it with zest.releaser 6.2? With that one, you shouldn't need twine on your command line. @mauritsvanrees fixed it last evening :-)

PR #143, opened 2015-10-29
Use twine as library.

mauritsvanrees:
This fixes issue #142.


reinout:
Every PR that removes as many (or more) lines than it adds is a good one :-)

LGTM, now that we're using twine, we should go all the way and use it as a library.

reinout:
I've released 6.2

Issue #155, opened 2016-01-08
Show last release and changelog prior to choose the next version


gforcada:
Background: I created [freitag.releaser](http://pypi.python.org/pypi/freitag.releaser) to be able to release packages in a bulk, it's basically a wrapper around zest.releaser.

Thus whenever I release some packages I never remember what was the last release version nor what where exactly the changes in each of the packages.

These two, changelog for the next release and the last release number, are the pieces of information I need to decide which version number I want to release.

So, would it be possible to show the changelog and the last version released before entering the next release number?

mauritsvanrees:
plone.releaser uses an entry point to show the changelog:
https://github.com/plone/plone.releaser/blob/957e402ef8a4f7a30c3190b305b8828d8726edc1/plone/releaser/release.py#L60

Maybe we could incorporate this in zest.releaser.

Showing the last version number: usually the recommended version is fine for me.  Do you often choose a different version?

Then again, it might be easy to do: if we show the changelog, we could show two lines more to include the next header and its underline in the file. That does need some care, as the code that we use for grabbing the history currently expects that the new version number has already been specified by the user.

mauritsvanrees:
I am trying it out now.  This shows when I do a prerelease of zest.releaser with that code (with minor edits for clarity):

```
$ prerelease 
Wrote entry point documentation to .../zest.releaser/doc/source/entrypoints.rst
Run pyroma on the package before tagging? (Y/n)? n
Do you want to run check-manifest? (Y/n)? n
Changelog entries for version 6.6:

6.6 (2016-01-08)
----------------

- Nothing changed yet.


6.5 (2016-01-05)
----------------
Enter version [6.6]: 
```

reinout:
Note that there's also "lasttaglog" that shows the (git) log messages since the last release. That might be too elaborate for your needs, though.

reinout:
With #156, I assume the issue is fixed?
(I'm going through old issues).

PR #156, opened 2016-01-08
Show changelog of current release before asking for new version.

mauritsvanrees:
Issue #155.

Implementation: we grab the version and the history twice:

- Initially in read-only mode: check current version and history and
  maybe warn about problems.

- Then show the changelog of the current release as found in the
  history.  Include the header and underline of the previous release
  for reference.

- Then grab the version in write mode: ask for the new version number.

- Then grab the history in write mode: take over the new version number.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,9 @@ Changelog for zest.releaser
 6.6 (unreleased)
 ----------------
 
-- Nothing changed yet.
+- Show changelog of current release before asking for the new version
+  number.  Issue #155.  [maurits]
+
 
 
 6.5 (2016-01-05)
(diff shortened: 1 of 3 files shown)


mauritsvanrees:
@reinout Do you agree with the strategy here, of basically grabbing the history and version twice? I could rework it in a different way, but this suggested itself to me as an easy way.

reinout:
Grabbing it twice: fine with me.

Printing even *more* output is something that might need  a second thought. I probably think it is OK as printing the changelog since the previous version sounds pretty useful.

OTOH, for projects where you almost always just increment the version number... for those it is overkill.

mauritsvanrees:
I still like this. @reinout did you have a second thought?

reinout:
"I still like this" from you is reason enough for me. I'll merge it :-) I assume the patch is still correct. I'll leave it to you to make the release.

Issue #160, opened 2016-01-28
Add bumpversion command


mauritsvanrees:
Background is again batch editing zope packages, just like issue #159:
https://mail.zope.org/pipermail/zope-dev/2016-January/047076.html

The idea would be to bump the version and update the changelog without doing the rest of the prerelease part. Important:

- Add an option for increasing the micro or minor or major version, to bump the respective version number part.
- Add an option for *keeping* the development part.

This is for cases where you know the next release needs to be more than a micro version number increase, but you do not want to release just yet. For example, you make a backwards incompatible change, but you are not the person who handles the releases.

Again, we will ask for confirmation: the version we display is a proposal for the user.  With `--no-input` you can accept the default without confirmation.

Examples:

- 1.2.3.dev0 becomes 1.2.3
- 1.2.3 with dev becomes 1.2.3.dev0
- 1.2.3.dev0 with --dev probably stays 1.2.3.dev0
- 1.2.3.dev0 with --minor becomes 1.3.0
- 1.2.3.dev0 with --dev and --minor becomes 1.3.0.dev0
- 1.3.0 with --dev becomes 1.3.0.dev0

Future idea (but maybe easy to add immediately): add options for alpha, beta, rc, final release.  So:

- 1.0.dev0 with --alpha becomes 1.0a1
- 1.0.dev0 with --alpha --dev becomes 1.0a1.dev0
- 1.0a1 with --final becomes 1.0

mauritsvanrees:
@mgedmin Would this help you?

reinout:
If something is 1.3.0 dev, 1.3.0 is not yet released. You're making a some change via an external script (the python 3.2 marker stuff from the mailinglist). Fine, something changed. Why would you want to remove the dev marker for that? And why would you want to turn it into a 1.3.1 (thus skipping the not-yet-released 1.3.0 release)?

I can imagine you have several packages at 1.3.2dev, 1.4.5dev and 1.1dev. You then want those to be turned into 1.3.0dev, 1.5.0dev and 1.1dev (this one stays the same?) if you run it with `--minor`, I guess.

Would that not be more logical?

I suspect a `--major` is also needed?

Reinout

mgedmin:
Here's my usecase:
- I release software version 1.2.3
- the version number in git master is incremented to 1.2.4.dev0 as part of the release process (thank you, zest.releaser, for automating this for me!)
- time passes
[...]

PR #161, opened 2016-01-28
Added addchangelogentry command

mauritsvanrees:
This implements issue #159.

This also moves some more methods around, so we reuse more code.

PR #162, opened 2016-01-29
Added bump version command

mauritsvanrees:
This implements issue #160.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,11 @@ Changelog for zest.releaser
 6.6 (unreleased)
 ----------------
 
+- Added ``bumpversion`` command.  Options ``--feature`` and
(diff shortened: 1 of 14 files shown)


mauritsvanrees:
Note that this contains the changes done in the open pull request #161 as it needs some of those changes.

reinout:
Looks fine in principle.

I'd add a note in the changelog to say that the exact functionality of bumpversion might change in the next version. Once you start using it for real I suspect there might be one or two small changes.

reinout:
I'll leave merging up to you, too many open intertwined PRs right now.

mauritsvanrees:
I have merged all and released it. I picked version 6.6.0 and set the next version to 6.6.1. So we start using semantic versions ourselves.  Let's see if we like it.

PR #174, opened 2016-04-21
Enable 3.5 tests

graingert:


reinout:
Good one!
I assume we can remove 3.3?

graingert:
@reinout nope, 3.3 is still supported until 2017-09-29 https://docs.python.org/devguide/index.html#branchstatus

reinout:
I meant it more like "almost nobody uses it anymore, so why waste travis-ci.org resources" :-) But of course we can keep it.

Thanks for adding 3.5!

PR #176, opened 2016-05-05
Support overriding the default development marker in setup.cfg

drucci:
This allows for overriding the default development marker of .dev0 with one specified in setup.cfg
within the zest.releaser section.

There are probably some additional changes that need to be made (haven't touched docs or other 
related files). Figured I'd see earlier on if this is something that would be considered added to zest.releaser.

It didn't seem like this is something that was do-able as a plugin.

### DIFF
--- a/zest/releaser/postrelease.py
+++ b/zest/releaser/postrelease.py
@@ -15,7 +15,7 @@ TODAY = datetime.datetime.today().strftime('%Y-%m-%d')
 HISTORY_HEADER = '%(new_version)s (unreleased)'
 NOTHING_CHANGED_YET = '- Nothing changed yet.'
 COMMIT_MSG = 'Back to development: %(new_version)s'
-DEV_VERSION_TEMPLATE = '%(new_version)s.dev0'
+DEV_VERSION_TEMPLATE = '%(new_version)s%(development_marker)s'
 
 DATA = {
     # Documentation for self.data.  You get runtime warnings when something is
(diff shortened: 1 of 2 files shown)


reinout:
Looks useful in principle. The `.dev0` marker is quite python-specific. As I myself also use it for arduino libraries, qgis plugins, documentation, etc and `.dev0` looks a bit funny there :-)

Where do you want to use it for?

drucci:
We use the suffix `-SNAPSHOT` for our development builds. Eventually they are published to our own instance of http://www.sonatype.org/nexus/ (It's more java oriented, hence the convention of using -SNAPSHOT).

drucci:
After using this a bit, I see there would be some other changes needed. suggest_version in utils checks for .dev0 specifically, so that would need to take into account the overridden one.

Also I notice postrelease seems to always bump the Z in X.Y.Z{devmarker} . Have you ever thought about having the --feature / --breaking functionality in postrelease?

mauritsvanrees:
I have taken over your changes in a new pull request #205. See explanation there. Thanks for your work!

Now your question about postrelease. I don't see the use of adding those command line options to `postrelease` (and `fullrelease`). Doing a postrelease or a feature bump are really two different things, best served by separate commands, in my opinion. Combining them would make the code harder to understand.
But maybe pull request #204 gives you some extra options. (I included that in the new pull request #205 to avoid merge conflicts later on.)

PR #178, opened 2016-06-13
Updated tox.ini and docs to reflect testing with Python 3.5

taschini:
Python 3.5 has been used for testing by Travis since the 21st of April 2016. This PR updates the docs to reflect this and adds a py35 environment to ``tox.ini`` too.


reinout:
Thanks!

PR #179, opened 2016-06-27
Add push-changes option to modify push default behavior

newlog:
The push-changes option can be specified in the setup.cfg
file under the [zest.releaser] section. This is a boolean
value. If set to yes, it set to 'yes' the default answer
regarding pushing version changes to remote in the post-release
stage. Otherwise, if set to 'no'. If not added to the configuration
file, the old behavior is honored, setting the default behavior
to 'yes'.

This change is useful if you run zest with the no-input option but
you do not want to push your changes, given that you might not have
a remote already specified.


mauritsvanrees:
Seems okay.
Can you add an entry in the change log and add a few lines to the documentation?

mauritsvanrees:
I have added a changelog entry and a few lines of documentation and have merged this manually.
Thank you!

newlog:
Thanks and sorry for not finishing the work.

PR #182, opened 2016-07-01
Mentions support for a CHANGES file with extension .md in README

swinkels:
zest.releaser recognizes a CHANGES file with extension md. The README only mentioned extensions rst, txt, and markdown, or no extension at all. I could not imagine .md was not recognized so I looked through the code and see there, it is supported. To avoid other developers having to do the same, I updated the README.


mauritsvanrees:
Merged, thanks.

PR #185, opened 2016-08-16
Less misleading description of what it does

mgedmin:
- zest.releaser hasn't been using `--formats=zip` since 2015, see #215.

- zest.release isn't *actually* using `setup.py upload` for the upload either, so mention `twine` briefly.


mauritsvanrees:
Thanks.
We are not using `setup.py register` either, so it looks like we should update this a bit more. I have created a new pull request #186 to reflect the above changes plus some extra. Can you have a look?
I will close your pull request.

PR #186, opened 2016-08-18
Updated the documentation on uploading.

mauritsvanrees:
We are not using `setup.py register/upload` anymore, but use twine for both.
Removed the old-style `.pypirc` from the docs.

This builds on and replaces pull request #185

### DIFF
--- a/doc/source/uploading.rst
+++ b/doc/source/uploading.rst
@@ -2,43 +2,41 @@ Uploading to pypi (or custom servers)
 =======================================
 
 When the (full)release command tries to upload your package to a pypi server,
-zest.releaser basically just executes the command ``python setup.py register
-sdist --formats=zip upload``.
+zest.releaser basically executes the command ``python setup.py sdist`` and does a
+``twine upload``.  The twine command replaces the less safe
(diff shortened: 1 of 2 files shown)


doc/source/uploading.rst
```diff
@@ -2,43 +2,40 @@ Uploading to pypi (or custom servers)
 =======================================
 
 When the (full)release command tries to upload your package to a pypi server,
-zest.releaser basically just executes the command ``python setup.py register
-sdist --formats=zip upload``.
+zest.releaser basically just executes the command ``twine upload``.
```
mgedmin: What about `python setup.py sdist`?  Does `twine upload` do that nowadays?

mgedmin: To be more constructive: perhaps rephrase as

> basically just executes the commands `python setup.py sdist` and `twine upload`.

mauritsvanrees: Ah, true, `setup.py sdist` is still needed yes. I have updated the text along those lines.

mgedmin:
LGTM!

Issue #183, opened 2016-07-06
zest.releaser does not work with newest twine


gforcada:
Hi,

I've been making some releases lately and before doing that I'm usually doing ``pip install -U pip zest.releaser[recommended]``.

This means that newer versions of twine were being installed (1.7+) and I had to revert them to version 1.6.5 (did not test other releases).

Unfortunately they don't provide a changelog file whatsoever, but after 1.6.5 already comes 1.7 and 1.7.1 and 1.7.2.

The traceback that I get is the following one:

```
There were errors or warnings.
ERROR: Package upload has failed.
Traceback (most recent call last):
  File "/haproxy_log_analysis/lib/python2.7/site-packages/zest/releaser/utils.py", line 749, in retry_twine
    return twine_function(*twine_args)
TypeError: upload() takes exactly 12 arguments (10 given)
```

reinout:
Their changelog is in the docs/ directory: https://github.com/pypa/twine/blob/79e0c8fe1991a6a613ff1ace107b646a6832dd5a/docs/changelog.rst

reinout:
The change that broke us: https://github.com/pypa/twine/commit/ece3711fb7e69891b1ed8f86314a2762dbe64454

We *are* calling a function with 10 parameters (which are now 12). That's not really a proper API function to call. Either we're calling the wrong function or they need to provide a good one or they ought to use keyword arguments.

reinout:
See https://github.com/pypa/twine/issues/194

sigmavirus24:
To summarize the discussion on twine, you *are* calling the wrong function. I've outlined roughly what you need to do to use our semi-public API.

grahamar:
Same with the register command

jensens:
Seems like twine is not really the right tool, overall sounds like its not meant to be used as API. What a shame, it comes in handy.

Anyway, pinning down to twine 1.6.5 and all works for now.

sigmavirus24:
> Seems like twine is not really the right tool, overall sounds like its not meant to be used as API.

@jensens as I explained to @reinout there's an API, it just isn't documented (hence why I referred to it as "semi-public"). I provided an example. If that is still not the right tool, then I'm sorry but I don't understand what you need.

mauritsvanrees:
`twine` is good to use. Options currently:

1. Let `zest.releaser` call `twine` on the command line instead of importing it and calling a function (`upload` or `register`).
[...]

PR #188, opened 2016-09-12
Support and require twine 1.8.0 as minimum version.

mauritsvanrees:
Fixes https://github.com/zestsoftware/zest.releaser/issues/183

Note: this means that there actually is no version of zest.releaser that works with twine 1.7.x. But that's okay.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,10 @@ Changelog for zest.releaser
 6.6.5 (unreleased)
 ------------------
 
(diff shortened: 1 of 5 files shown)


reinout:
Ah, right, yes, *that* issue :-)

reinout:
You're going to do a release later on?

mauritsvanrees:
I have released 6.6.5, using twine 1.8.1. :-)

PR #189, opened 2016-09-13
Use the intended API of twine.

mauritsvanrees:
This should work with twine 1.6.0 and higher, including future versions.

We were using the `upload` and `register` functions, which were never meant to be used by third party code. The needed arguments to those versions can and have changed in backwards incompatible ways, so we should not rely on them.  And if you dare to suggest to make those functions more future proof and stable by using [keyword arguments](https://github.com/pypa/twine/pull/212), you'd better put on your asbestos underwear: one of the authors threatens to intentionally break the use of these functions...

So we need to rewrite a bunch of code. So be it. At least the intended API is decent. But it's a pity that we need to read more from the PyPI config now (repository url, username, password). And you can no longer pass a list of filenames (sdist plus wheel usually) for uploading, but need to upload them separately. This needed a rewrite of the way we handle retries, mostly because for retries we need to read the pypiconfig fresh: the user may have edited it to solve an upload problem. Can't really blame the API for that.


mauritsvanrees:
@reinout Could you have a look at this?
Greetings from Plone conference Boston. :-)

reinout:
Looks fine, I'll merge it.

You'll do a release? I don't know if you've got other stuff you want to get in?

mauritsvanrees:
While in Boston airport, I have released 6.7 with this. :-)

PR #192, opened 2016-11-07
Modified zest.releaser to avoid "register" command when not publishing to public PyPi

willowmck:
The private PyPi cloud repositories often don’t use the register
command causing publish to fail.  Don’t use this command unless you are
publishing to the public one.

### DIFF
--- a/zest/releaser/pypi.py
+++ b/zest/releaser/pypi.py
@@ -247,14 +247,7 @@ class PypiConfig(object):
         mixed case and specify 0, false, no or off for boolean False,
         and 1, on, true or yes for boolean True.
         """
-        default = True
-        if self.config is None:
-            return default
-        try:
-            result = self.config.getboolean('zest.releaser', 'release')
-        except (NoSectionError, NoOptionError, ValueError):
-            return default
-        return result
+        return self.__get_boolean('zest.releaser', 'release')
 
     def extra_message(self):
         """Return extra text to be added to commit messages.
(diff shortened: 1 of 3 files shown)


mauritsvanrees:
Actually, on the public pypi, separately registering is [not needed anymore](https://github.com/pypa/twine/issues/200) and will actually fail with current twine.

The public pypi is also the only server where we know for sure how to ask if a package is already registered or not. But this is no longer needed, because, as said, registering is not needed.

So we need to clean this up. A proposal:

- No longer check if a package is registered on PyPI (or anywhere else, but we were not doing that anyway).

- Split the 'do you want to register/upload' question into two, for clarity.

- If the repository url is https://upload.pypi.org/legacy/ (or maybe: starts with `https://upload.pypi.org`) don't even ask to register, as it is not needed and not supported.

- For other repositories, do ask if the user wants to register.

- For the default answer, we read a new `register` boolean config option from setup.cfg/pypirc. By default: don't register.

Does that sound sane?

It it quite different from the approach in your pull request. Would you be willing to work on this new approach? Otherwise I can see when I have the time.

willowmck:
Yes, that sounds like a good approach.  I can probably get to this work later this week or early next week.  Or, if you feel like doing it, that's fine too.

willowmck:
Hi, I'm thinking that the URL equivalence check is not necessary if we simply turn off register by default and make it a configuration option.  Does that approach sound okay to you?

zest/releaser/pypi.py
```diff
@@ -316,11 +315,13 @@ def no_input(self):
...
-        default = False
-        if self.config is None:
-            return default
-        try:
-            result = self.config.getboolean('zest.releaser', 'no-input')
-        except (NoSectionError, NoOptionError, ValueError):
-            return default
+        return self.__get_boolean('zest.releaser', 'no-input')
+
+    def __get_boolean(self, section, key):
```
mauritsvanrees: The `__get_boolean` method should accept a default value. Currently you make the default False. In `want_release` we want True as default, which is not happening now.

Also, I would rename this to `_get_boolean` or `_getboolean`. A double underscore indicates a special method, which is not the case here.

zest/releaser/release.py
```diff
@@ -144,35 +141,15 @@ def _upload_distributions(self, package):
...
-                else:
-                    # We must register first.
-                    do_register = True
-                    question = "Register and upload"
-            else:
-                # Assume register unnecessary unless publishing to Public PyPi.
-                do_register = False
-                question = "Upload"
+            if self.pypiconfig.register_package():
+                logger.info("Registering...")
```
mauritsvanrees: If a user has multiple servers in his `~.pypirc`, this change means we register at all three servers, without asking any questions.
I would prefer this:
- Ask if user wants to upload to server X.
- No: don't upload and don't register.
- Yes: if register is on in the config, we register. And then we upload.

mauritsvanrees:
I have added a few remarks in the commit. For the rest it looks good.
Well, Travis complains. :-) And in the doctests it can be very tricky to see what the exact problem is. :-/

mauritsvanrees:
This was replaced by your own new pull request #194, right? I will close this one then.

PR #194, opened 2016-11-29
Update pypi to set a default value for booleans

willowmck:
Also rearranged register logic to avoid registering when the user does not wish to upload


mauritsvanrees:
Does this replace your pull request #192 ?

There are test failures. Looks like the problem is near line 83 in `functional-git.txt` where the expected output still contains `MOCK twine register`.

mauritsvanrees:
Also, did you notice my comment on your previous pull request? See https://github.com/zestsoftware/zest.releaser/pull/192#discussion_r88517568

Actually, let me copy that remark here:

If a user has multiple servers in his `~.pypirc`, this change means we register at all three servers, without asking any questions.
I would prefer this:

- Ask if user wants to upload to server X.
- No: don't upload and don't register.
- Yes: if register is on in the config, we register. And then we upload.

willowmck:
Thanks for pointing out the test failures.  I tried to run the test locally which turned out to be a bad idea since it interacted with git on my behalf.  Probably not what I wanted.  So, it was hard to reproduce the exact error.  At any rate, I will address the problem you pointed out. 

As for the comments regarding uploading, I did make a change to only register if the user wishes to upload.  You should be able to see that in the release.py around lines 144-154.

```
        for server in servers:
            question = "Upload"
            default = True
            exact = False
            if utils.ask("%s to %s" % (question, server),
                         default=default, exact=exact):
                if self.pypiconfig.register_package():
                    logger.info("Registering...")
                    # We only need the first file, it has all the needed info
                    self._retry_twine('register', server, files_in_dist[0])
                for filename in files_in_dist:
                    self._retry_twine('upload', server, filename)
```

However, I didn't change the form of the question to include the server name, which is probably not enough information to the user so I'll make that change as well.

willowmck:
Thanks for pointing out the test failures.  I tried to run the tests locally, which turned out to be a not so great idea since it committed code to my git repository.  Not exactly what I wanted.

I think I misunderstood you the first time I read your comment on register/upload.  I'll make that change as well.

mauritsvanrees:
When I look at your commits on github, I see a question mark behind your name, where github indicates that it does not recognise your email address. You may want to have a look at that. I don't mind, it just struck my eye.

The tests should not make changes to existing git repositories when you run them locally. They just create temporary repositories and do their stuff there. So that sounds weird.

There is still a test failure. Instead of 

```
Question: Upload to pypi (y/N)?
```

we get:

```
Question: Upload to pypi (Y/n)?
```

So the default answer that we get is Yes instead of No.
I think we *do* want No here, as this particular test checks what happens when a package is not yet on PyPI. With No as default, this means you won't accidentally publish secret internal packages on PyPI if you unthinkingly press Enter on all questions.
I would like to keep that logic. I did not realise this when earlier I wrote in my proposal: 'No longer check if a package is registered on PyPI.'
So the intended logic is: if a package is not yet registered on PyPI, then by default don't register *and* don't upload. So the default answer for 'Upload to pypi' in that case should be No.
Does that still make sense?

It should be easy, just put back a few lines that you removed:

```
on_pypi = package_in_pypi(package)
...
          if server == 'pypi' and not on_pypi:
               logger.info("This package is NOT registered on PyPI.")
              # We are not yet on pypi.  To avoid an 'Oops...,
              # sorry!' when registering and uploading an internal
              # package we default to False here.
              default = False
              exact = True
              if utils.ask("%s to %s" % (question, server),
                           default=default, exact=exact):
```

Ah, you are right about the multiple servers.
One thing: for a slight speed increase I would calculate `self.pypiconfig.register_package()` only once. So something like this:

```
    register = self.pypiconfig.register_package()
    for server in servers:
        ...
            if register:
                logger.info("Registering...")
```

mauritsvanrees:
This needed a rebase due to an unrelated pull request that I merged last week, plus I wanted a few small changes, and some documentation. So I created pull request #200 based on your commits and a few extra.

Note: most important change for you would be that the option is now `[zest.releaser] register`.

I will close this one. I plan to merge my one soon. Thanks for your work!

PR #197, opened 2016-12-21
Allow `before_upload` hook to create dists

t-8ch:
Create the list of distributions after the before_upload hook has fired.
This allows the before_upload hook to create additional distributions,
which will then be uploaded.

### DIFF
--- a/zest/releaser/release.py
+++ b/zest/releaser/release.py
@@ -126,15 +126,15 @@ class Releaser(baserelease.Basereleaser):
                         pypi.DIST_CONFIG_FILE)
             return
 
-        # Get list of all files to upload.
-        files_in_dist = [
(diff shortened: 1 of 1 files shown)


mauritsvanrees:
Seems good to me.
The tests fail because of an error in the buildout phase. I will create a different pull request to address that, and then we can try your pull request again.

mauritsvanrees:
Can you rebase on master? That should fix the bootstrap problem on Travis, so we can see if any real problems surface.

t-8ch:
@mauritsvanrees 
Rebased, pushed, and test are passing

mauritsvanrees:
Perfect, thank you! I'll merge.

Oh, it is always helpful to add a line to `CHANGES.rst`. I will do that myself.

mauritsvanrees:
I have released 6.7.1 with this.

PR #204, opened 2016-12-30
Added version-levels and less-zeroes options.

mauritsvanrees:
This influences the suggested version.  The defaults give the same results as previously. I tested this by copying the `utils.txt` tests from this branch, removing the lines that use the new options, and running them with the code from the master branch.

I myself am going to prefer `less-zeroes = yes` and `version-levels = 3` in my `~/.pypirc`, which gives the results that the Plone Release Team prefers.

The options are documented in `options.rst`, copied here for good measure:

**less-zeroes = yes / no**
Default: no.
This influences the version suggested by the bumpversion command.
When set to true:

- Instead of 1.3.0 we will suggest 1.3.
- Instead of 2.0.0 we will suggest 2.0.

**version-levels = a number**
Default: 0.
This influences the version suggested by the postrelease and bumpversion commands.
The default of zero means: no preference, so use the length of the current number.

This means when suggesting a next version after 1.2:

- with 0 we will suggest 1.3: no change in length
- with 1 we will still suggest 1.3, as we will not
 use this to remove numbers, only to add them
- with 2 we will suggest 1.3
- with 3 we will suggest 1.2.1

If the current version number has more levels, we keep them.
So with ``version-levels=1`` the next version for 1.2.3.4 will be 1.2.3.5.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,9 @@ Changelog for zest.releaser
 6.8 (unreleased)
 ----------------
 
+- Added ``version-levels`` and ``less-zeroes`` options.
+  This influences the suggested version.  [maurits]
+
 - Allow ``.pypirc`` with just a ``pypi`` section.  Previously, we
   required either a ``[server-login]`` section with a ``username``
   option, or a ``[distutils]`` section with an ``index-servers`` option.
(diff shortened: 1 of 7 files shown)


mauritsvanrees:
No, but I am part of the Plone Release Team, which assists the release manager by making releases of individual packages. :-)
Thanks for the reviews!

PR #209, opened 2017-02-13
Add tag formatter in the config

### LINKED ISSUE #157: Add a config for tag-prefix
We are tagging our changes with the prefix `v`. zest.releaser should be be allow specifying this.

tcezard:
This PR adds a tag formatter in the `.pypirc` options  and should resolve #157
Do let me know if I missed something and am happy to change it.

### DIFF
--- a/doc/source/options.rst
+++ b/doc/source/options.rst
@@ -132,6 +132,11 @@ development-marker = a string
     This is the development marker.
     This is what gets appended to the version in postrelease.
 
+tag-format = a string
+    Default: ``%(version)s``
+    This is a formatter that change the name of the tag.
+    It needs to contain ``%(version)s``
+
 
 Per project options
 -------------------
--- a/doc/source/options.rst
+++ b/doc/source/options.rst
@@ -134,7 +134,7 @@ development-marker = a string
 
 tag-format = a string
     Default: ``%(version)s``
-    This is a formatter that change the name of the tag.
+    This is a formatter that changes the name of the tag.
     It needs to contain ``%(version)s``
 
 
--- a/zest/releaser/pypi.py
+++ b/zest/releaser/pypi.py
@@ -426,13 +426,12 @@ class PypiConfig(object):
         try:
             result = self.config.get(
                 'zest.releaser', 'tag-format')
-            #test the formater
-            if not re.search('%\(version\)s', result):
-                print('%(version)s needs to be part of the string formater')
-                sys.exit(1)
         except (NoSectionError, NoOptionError, ValueError):
             return default
-
+        # test the formatter
+        if not re.search('%\(version\)s', result):
+            print('%(version)s needs to be part of the string formatter')
+            sys.exit(1)
         return result
 
     def _get_boolean(self, section, key, default=False):
--- a/zest/releaser/pypi.py
+++ b/zest/releaser/pypi.py
@@ -407,6 +407,31 @@ class PypiConfig(object):
             return default
         return result
 
+    def tag_format(self):
+        """Return the string formating the tag name in the release.
+
+        Override the default ``%(version)`` in setup.cfg using
+        a ``tag-format`` option::
+
+            [zest.releaser]
+            tag-format = v%(version)s
+
+        Returns default of ``%(version)s`` when nothing has been configured.
+        """
+        default = '%(version)s'
+        if self.config is None:
+            return default
+        try:
+            result = self.config.get(
+                'zest.releaser', 'tag-format')
+            try:
+                '%(version)s'%({'version':''})
+            except KeyError:
+                raise ValueError('version needs to be part of the string formater use default')
+        except (NoSectionError, NoOptionError, ValueError):
+            return default
+        return result
+
     def _get_boolean(self, section, key, default=False):
         """Get a boolean from the config.
 
--- a/zest/releaser/pypi.py
+++ b/zest/releaser/pypi.py
@@ -4,6 +4,7 @@ import codecs
 import logging
 import os
 import sys
+import re
 
 import pkg_resources
 from six.moves.configparser import ConfigParser
@@ -426,13 +427,12 @@ class PypiConfig(object):
             result = self.config.get(
                 'zest.releaser', 'tag-format')
             #test the formater
-            result % ({'version':''})
-
+            if not re.search('%\(version\)s', result):
+                print('%(version)s needs to be part of the string formater')
+                sys.exit(1)
         except (NoSectionError, NoOptionError, ValueError):
             return default
-        except KeyError:
-            print('%(version)s needs to be part of the string formater')
-            sys.exit(1)
+
         return result
 
     def _get_boolean(self, section, key, default=False):
--- a/zest/releaser/pypi.py
+++ b/zest/releaser/pypi.py
@@ -3,6 +3,7 @@ from __future__ import unicode_literals
 import codecs
 import logging
 import os
+import sys
 
 import pkg_resources
 from six.moves.configparser import ConfigParser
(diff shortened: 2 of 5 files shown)


doc/source/options.rst
```diff
@@ -132,6 +132,11 @@ development-marker = a string
     This is the development marker.
     This is what gets appended to the version in postrelease.
 
+tag-format = a string
+    Default: ``%(version)s``
+    This is a formatter that change the name of the tag.
```
reinout: change => changes

zest/releaser/pypi.py
```diff
@@ -407,6 +407,31 @@ def version_levels(self):
             return default
         return result
 
+    def tag_format(self):
+        """Return the string formating the tag name in the release.
+
+        Override the default ``%(version)`` in setup.cfg using
```
reinout: Missing `s` behind `%(version)`

zest/releaser/pypi.py
```diff
@@ -407,6 +407,31 @@ def version_levels(self):
             return default
         return result
 
+    def tag_format(self):
+        """Return the string formating the tag name in the release.
```
reinout: "the string formating the tag name" => "the string format for the tag name"

zest/releaser/pypi.py
```diff
@@ -407,6 +407,31 @@ def version_levels(self):
...
+
+            [zest.releaser]
+            tag-format = v%(version)s
+
+        Returns default of ``%(version)s`` when nothing has been configured.
+        """
+        default = '%(version)s'
+        if self.config is None:
+            return default
+        try:
```
reinout: Better unwind the two try/excepts. So a "try/except NoSectionError" (with `return default`) and then on the same level the "try/except KeyError".

zest/releaser/pypi.py
```diff
@@ -407,6 +407,31 @@ def version_levels(self):
...
+            result = self.config.get(
+                'zest.releaser', 'tag-format')
+            try:
+                '%(version)s'%({'version':''})
```
reinout: pep8: spaces around `%`

zest/releaser/pypi.py
```diff
@@ -407,6 +407,31 @@ def version_levels(self):
...
+            except KeyError:
+                raise ValueError('version needs to be part of the string formater use default')
```
reinout: `%(version)s needs to be part of the string format`

tcezard: I realised here that the `ValueError` will get caught and not reported.
Is it ok to silently fallback on the default if the tag format is missing the version?
Should I `print()` a warning or should we `exit` completely?

reinout: See my comment earlier by not nesting the two try/excepts. That immediately removes this problem.

But... it is good to print the error and do a `sys.exit(1).

reinout: Note: check the rest of the code on how we print something before a "sys.exit", I don't remember if we use logging or `print()` right now :-)

zest/releaser/pypi.py
```diff
@@ -407,6 +408,33 @@ def version_levels(self):
...
+        Returns default of ``%(version)s`` when nothing has been configured.
+        """
+        default = '%(version)s'
+        if self.config is None:
+            return default
+        try:
+            result = self.config.get(
+                'zest.releaser', 'tag-format')
+            #test the formater
+            result % ({'version':''})
```
mauritsvanrees: This does not actually catch an error. If the format is `foo` the code above won't complain:

```
>>> 'foo' % {'version': ''}
'foo'
```

Trying it out is a good idea, but currently it doesn't help. I don't really know a better way at the moment.

zest/releaser/release.py
```diff
@@ -98,10 +100,10 @@ def _make_tag(self):
             else:
                 # all commands are needed in order to proceed normally
                 print("Please create a tag for %s yourself and rerun." %
-                      (self.data['version'],))
+                      (version,))
```
mauritsvanrees: I would pass `tag` here instead of `version`, to remind the user what the tag format normally is.
I don't think I ever ran into this part of the code though.

zest/releaser/release.py
```diff
@@ -251,9 +253,9 @@ def _release(self):
             return
 
         package = self.vcs.name
-        version = self.data['version']
+        tag = self.pypiconfig.tag_format() % ({'version': self.data['version']})
         logger.info("Doing a checkout...")
-        self.vcs.checkout_from_tag(version)
+        self.vcs.checkout_from_tag(tag)
         # ^^^ This changes directory to a temp folder.
         self.data['tagdir'] = os.path.realpath(os.getcwd())
```
mauritsvanrees: To assist possible zest.releaser extensions that want to do something with tags, it would be good to store it in the data:

```
self.data['tag'] = tag
```

Then you would need to add a line of documentation in the [`DATA`](https://github.com/zestsoftware/zest.releaser/blob/6.8.1/zest/releaser/release.py#L24) dictionary near the top of the file, something like: `'tag': "Tag we're releasing",`.

zest/releaser/pypi.py
tcezard: That is right. Thanks for spotting that.
I was trying to protect against typo like `'%(verion)s'` which I would definitely do.
Only way to  would be to use regex like:
```python
re.search('%\(version\)s', result)
```

zest/releaser/release.py
tcezard: I added this in the prepare function
```python
    def prepare(self):
        """Collect some data needed for releasing"""
        self._grab_version()
        self.data['tag'] = self.pypiconfig.tag_format() % (
            {'version': self.data['version']})
        self._check_if_tag_already_exists()
```
But the tests are now failing because they do not run `prepare`
Should I store the tag in `_grab_version` or in another `_grab...` function in `Basereleaser` and call it in the test?

mauritsvanrees: Ah, I see. It is probably easiest and best to change the tests in `release.txt` to no longer call `_grab_version` and `_check_if_tag_already_exists` individually, but instead to call `prepare`.

zest/releaser/pypi.py
mauritsvanrees: I agree. So something like this:

```
try:
    result = self.config.get('zest.releaser', 'tag-format')
except (NoSectionError, NoOptionError, ValueError):
    return default
# And then continue with the checking.
```

BTW, `formater` should be spelled with two t's: `formatter`.

zest/releaser/pypi.py
mauritsvanrees: That indeed seems helpful.

tcezard:
I believe this PR is now addressing the different comments.
Let me know if I forgot anything

zest/releaser/pypi.py
```diff
@@ -407,6 +409,31 @@ def version_levels(self):
...
+        if self.config is None:
+            return default
+        try:
+            result = self.config.get(
+                'zest.releaser', 'tag-format')
+        except (NoSectionError, NoOptionError, ValueError):
+            return default
+        # test the formatter
+        if not re.search('%\(version\)s', result):
+            print('%(version)s needs to be part of the string formatter')
```
reinout: One small remaining tweak: be more specific about what *exactly* is wrong ('tag-format') and print the current value (in case one config file overrides another and you're left wondering what's wrong :-) ):

  print("%%(version)s needs to be part of 'tag-format': %s" % result)

I hope I have escaped the first `%` in the right way.

reinout:
@tcezard looks fine. I've added one extra suggestion (https://github.com/zestsoftware/zest.releaser/pull/209#pullrequestreview-22444514) that ought to make the error message clearer. (I can do that one myself, too, but I'm waiting for @mauritsvanrees to have a last look).

tcezard:
Thanks for all the helpful comments!

mauritsvanrees:
Thanks for your contribution!
I have released this as version 6.9.

PR #211, opened 2017-02-15
Prevent a crash in case the output of the command is not a valid utf-8 sequence.

### LINKED ISSUE #210: git diff encoding different from UTF-8
If you create a file in latin-1, commit it and then change its encoding in utf-8 before running prerelease for instance you will get a UnicodeDecode error.

```
  File "/usr/local/bin/prerelease", line 11, in <module>
    sys.exit(main())
  File "/usr/local/lib/python2.7/dist-packages/zest/releaser/prerelease.py", line 103, in main
    prereleaser.run()
  File "/usr/local/lib/python2.7/dist-packages/zest/releaser/baserelease.py", line 353, in run
    self.execute()
  File "/usr/local/lib/python2.7/dist-packages/zest/releaser/prerelease.py", line 69, in execute
    self._diff_and_commit()
  File "/usr/local/lib/python2.7/dist-packages/zest/releaser/baserelease.py", line 316, in _diff_and_commit
    diff = execute_command(diff_cmd)
  File "/usr/local/lib/python2.7/dist-packages/zest/releaser/utils.py", line 771, in execute_command
    result = _execute_command(command)
  File "/usr/local/lib/python2.7/dist-packages/zest/releaser/utils.py", line 698, in _execute_command
    stdout_output = stdout_output.decode(OUTPUT_ENCODING)
  File "/usr/lib/python2.7/encodings/utf_8.py", line 16, in decode
    return codecs.utf_8_decode(input, errors, True)
UnicodeDecodeError: 'utf8' codec can't decode byte 0xa7 in position 4197: invalid start byte
```

I guess it is a edge case of me using prerelease before having commited this change but I guess that's something we could prevent.

Natim:
Fixes #210

### DIFF
--- a/zest/releaser/utils.py
+++ b/zest/releaser/utils.py
@@ -695,9 +695,17 @@ def _execute_command(command, input_value=''):
     stderr_output = e.read()
     # We assume that the output from commands we're running is text.
     if not isinstance(stdout_output, six.text_type):
-        stdout_output = stdout_output.decode(OUTPUT_ENCODING)
+        try:
+            stdout_output = stdout_output.decode(OUTPUT_ENCODING)
+        except UnicodeDecodeError:
+            stdout_output = '%s stdout output is not a valid %s stream.' % (
+                command, OUTPUT_ENCODING)
     if not isinstance(stderr_output, six.text_type):
-        stderr_output = stderr_output.decode(OUTPUT_ENCODING)
+        try:
+            stderr_output = stderr_output.decode(OUTPUT_ENCODING)
+        except UnicodeDecodeError:
+            stderr_output = '%s stderr output is not a valid %s stream.' % (
+                command, OUTPUT_ENCODING)
     # TODO.  Note that the returncode is always None, also after
     # running p.kill().  The shell=True may be tripping us up.  For
     # some ideas, see http://stackoverflow.com/questions/4789837


zest/releaser/utils.py
```diff
@@ -695,9 +695,17 @@ def _execute_command(command, input_value=''):
     stderr_output = e.read()
     # We assume that the output from commands we're running is text.
     if not isinstance(stdout_output, six.text_type):
-        stdout_output = stdout_output.decode(OUTPUT_ENCODING)
+        try:
+            stdout_output = stdout_output.decode(OUTPUT_ENCODING)
+        except UnicodeDecodeError:
+            stdout_output = '%s stdout output is not a valid %s stream.' % (
```
reinout: So... instead of getting back output from, for instance, "git diff" we now get an error message?

This silences the error: is this **always** a good thing?

Natim: Another way to handle that would be not to decode the output.

Since we are printing it back we could just keep it as bytes instead so that the terminal whould have to handle the display and it would look really close to the original one.

Natim: > Is it a good thing to silence the error?

To be honest I don't think we should crash on UnicodeDecodeError then my solution might not be the right one, we should talk about it and see how we want to handle it.

mauritsvanrees: We could do `.decode('utf-8', 'replace')`.

mauritsvanrees:
If this is still a problem, please try with latest version [7.0.0a1](https://pypi.org/project/zest.releaser/7.0.0a1/). Python 3 only, so maybe encoding problems have gone away.

PR #215, opened 2017-04-10
Updated package versions in the test buildout

mauritsvanrees:


mauritsvanrees:
Ehmmm, I was waiting for Travis... :-) Which failed. I'll revert some of these. The new Sphinx fails on Python 3.3. And html5blib 0.999999999999999999999999 fails almost everywhere.

reinout:
Oopsie....

PR #221, opened 2017-05-18
Option to configure date format

mgedmin:
Adds date format in the config.  Default is ISO-8601 (%Y-%m-%d).

Put ``date-format = format string`` in your ``~/.pypirc`` or ``setup.cfg``.

My motivation for this is to maintain the historical changelog format of Dozer:
https://github.com/mgedmin/dozer/blob/master/CHANGELOG.rst


mgedmin:
The reason I didn't add any tests for the new code is that I looked how the `tag-format` option was being tested and found no tests for me to copy.  ;)

mauritsvanrees:
Thanks! I have released this as 6.12.

PR #222, opened 2017-06-07
If [bdist_wheel]universal is True, use that as the default for create_wheel

jamadden:
If users are specifying that they want the project to build universal wheels, then we should upload them by default because there is no downside.

This avoids having to repeat the same configuration twice, once for bdist_wheel and once for zest.releaser.

This was discussed at  https://github.com/zopefoundation/zope.annotation/pull/5#discussion_r120539701


jamadden:
There seems to be a problem in the buildout preventing Travis from running:
```
Version and requirements information containing requests-toolbelt:
  [versions] constraint on requests-toolbelt: 0.7.1
  Requirement of twine>=1.6.0: requests-toolbelt>=0.8.0
While:
  Installing script.
Error: The requirement ('requests-toolbelt>=0.8.0') is not allowed by your [versions] constraint (0.7.1)
```

I tested locally with tox and did not have this issue.

zest/releaser/pypi.py
```diff
@@ -277,12 +277,26 @@ def create_wheel(self):
...
+        universal = 1
+
         """
         if not USE_WHEEL:
             # If the wheel package is not available, we obviously
             # cannot create wheels.
             return False
-        return self._get_boolean('zest.releaser', 'create-wheel')
+        create_setting = self._get_boolean('zest.releaser', 'create-wheel', self)
+        if create_setting is not self:
```
mauritsvanrees: The comparison to `self` looks really weird to me. I realise what you are doing, but can you change this into comparison with `None` instead?

jamadden: Sure, done. I habitually use some sort of marker when I don't know if `None` is a possible return value, but it doesn't seem like it can be here.

mauritsvanrees:
Looking good. Merged. Thank you!

jamadden:
Thank you.

mauritsvanrees:
I have released 6.11 with this.

PR #228, opened 2017-06-21
Escape git clone path so it works when windows drives are mounted

halkeye:
I use the linux subsystem in windows all the time. It maps things to my windows username, which in this case happens to be "/mnt/c/Users/Gavin Mogan" so when the git clone happens, it thinks its multiple parameters.

Quote the path so it doesn't care about spaces.


mauritsvanrees:
A quick check tells me this can also happen on Linux or Mac when you have a space in your path.
The fix looks good. Thanks!

mauritsvanrees:
Your fix is included in release 6.12.1.

PR #230, opened 2017-07-14
[WIP] setup.cfg can also be a place to store the version

ewjoachim:
Also, the version line is not rewritten entirely (keeps comments etc)
Updated doc accordingly.

This is my first PR. I'm guessing it's a WIP because I didn't write tests, but I'd appreciate some guidance on how to write tests for this case.

I'm having a lot of test failures that seem unrelated to my own changes, so I'd like to see how it goes with the CI before spending time on those.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,7 @@ Changelog for zest.releaser
 6.12.3 (unreleased)
 -------------------
 
-- Nothing changed yet.
+- Write version in ``setup.cfg``. [ewjoachim]
 
 
 6.12.2 (2017-07-13)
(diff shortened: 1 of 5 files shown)


ewjoachim:
Hm :/

It seems most of the failures are because I tried to be smart, and replace the old version by the new one, instead of rebuilding the new line from scratch (but it seems overkill to rewrite it...)

Would it be ok to grab the regex pattern from `packaging`:  https://github.com/pypa/packaging/blob/16.7/packaging/version.py#L159-L188

That way we could replace "anything that looks like a version" in the line that's matching "version =" with our new version, and that should work with all cases.

Or I could just stick to the original way of doing it. There would be duplicated code, but...

Well any opinion would be appreciated ! :)

ewjoachim:
See #231 for a simpler approach, with code duplication.

ewjoachim:
Ok, closing this one, the other one is preferred.

PR #231, opened 2017-07-15
setup.cfg can also be a place to store the version

ewjoachim:
This is another implementation of #230. I'm not really satisfied because this is more or less copy/pasted code. (If we decide to go this way, I'll integrate the docs in this PR too).

Let me know what you prefer :)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,7 @@ Changelog for zest.releaser
 6.12.4 (unreleased)
 -------------------
 
-- Nothing changed yet.
+- Also support version in setup.cfg. [ewjoachim]
 
 
 6.12.3 (2017-08-16)
--- a/doc/source/versions.rst
+++ b/doc/source/versions.rst
@@ -5,7 +5,7 @@ Where does the version come from?
 ---------------------------------
 
 A version number is essentially what zest.releaser cannot do without. A
-version number can come from three different locations:
+version number can come from four different locations:
 
 - The ``setup.py`` file. Two styles are supported::
 
@@ -20,6 +20,12 @@ version number can come from three different locations:
     def setup(
         version='1.0',
         name='...
+
+- The ``setup.cfg`` file. zest.releaser will look for something like::
+
+    [metadata]
+    name = ...
+    version = 1.0
 
 - If no ``setup.py`` is found, zest.releaser looks for a ``version.txt``
   file. It should contain just a version number (a newline at the end is OK).
@@ -45,7 +51,7 @@ version number can come from three different locations:
 Where is the version number being set?
 --------------------------------------
 
-Of those three locations where the version can come from, only the first one
+Of those four locations where the version can come from, only the first one
 found is also set to the new value again. Zest.releaser assumes that there's
 only *one* location.
 
@@ -54,8 +60,8 @@ only *one* location.
 have **one** source and all the others should be derived from it.
 
 
-Using the version number in ``setup.py`` and as ``__version__``
----------------------------------------------------------------
+Using the version number in ``setup.py`` or ``setup.cfg`` as ``__version__``
+----------------------------------------------------------------------------
 
 Here are opinionated suggestions from the zest.releaser main authors about how
 to use the version information. For some other ideas, see the `zest.releaser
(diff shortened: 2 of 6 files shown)


mauritsvanrees:
Both approaches look okay at first glance. This one seems simpler, with less changes. I have not tried it out. Apparently `setup.cfg` could also contain `version = attr: src.VERSION`. I don't know if that maybe trips up the logic here.
I don't really have time now for a good look, and will be away for a week. And I think @reinout is on vacation but could be back soon.

ewjoachim:
After re-reading myself, this version needs the change in the regex pattern too. This regex pattern includes a digit so we're safe for version = attrs:

I'm totally ok if we leave this PR sleep for a few weeks :) Enjoy your time off !

ewjoachim:
Ping @reinout ?

reinout:
This one seems simpler, so I prefer it.
But: I haven't got experience with version-in-setup.cfg . Does that have to do with the new non-setup.py initiatives?

ewjoachim:
Well, I'm not sure what the new non-setup.py initiatives are, but yes, the idea is mainly to put things that should be static configuration in static config files instead of code files.

For an example of how that goes, see Raincoat's [setup.py](https://github.com/novafloss/raincoat/blob/master/setup.py) and [setup.cfg](https://github.com/novafloss/raincoat/blob/master/setup.cfg). All of this is being made possible by setuptool's new (last december) [ability]((https://setuptools.readthedocs.io/en/latest/setuptools.html#configuring-setup-using-setup-cfg-files)) to derive all parameters (sadly, except entrypoint) from setup.cfg.

I'll add the docs from the other PR in this PR.

ewjoachim:
Ok, I've updated and rebased the code and docs. I feel like it's missing tests (well, I know it's missing tests, because I tried it on my project and it didn't work at first). I'd be happy to write it, but the existing tests are in a format that, while readable, is quite new to me, and I'm not sure where and how it would be the most appropriate for me to write those.

ewjoachim:
Ok, it's updated with tests for the new feature. Ready for review !

doc/source/versions.rst
```diff
@@ -21,6 +21,12 @@ version number can come from three different locations:
         version='1.0',
         name='...
 
+- The ``setup.cfg`` file. zest.releaser will look for something like:
```
reinout: Double colon instead of single one

zest/releaser/tests/vcs.txt
```diff
@@ -245,3 +245,18 @@ Setup.py with an uppercase VERSION:
...
     setup(name='urgh', version=VERSION)
+
+version in setup.cfg:
+
+    >>> setupcfgproject = os.path.join(tempdir, 'setupcfg')
+    >>> os.mkdir(setupcfgproject)
+    >>> os.chdir(setupcfgproject)
+    >>> lines = [
+    ...     "[metadata]",
+    ...     "  version = 1.0"]
```
reinout: Remove the spaces in front of "version", this breaks the configparser on python 2.7 and on pypy. (That is where the test failure on this PR comes from).

CHANGES.rst
```diff
@@ -4,7 +4,7 @@ Changelog for zest.releaser
 6.12.4 (unreleased)
 -------------------
 
-- Nothing changed yet.
+- Write version in ``setup.cfg``. [ewjoachim]
```
reinout: Oh, and change this in "Also support version in ``setup.cfg``. "

reinout:
Looks good. I've got a few small comments (including a fix for the test failure).

ewjoachim:
All fixed. As it was simple fixes, I've taken the step of squashing fixups directly.

Thank you again !

CHANGES.rst
```diff
@@ -5,6 +5,8 @@ Changelog for zest.releaser
 -------------------
 
 - Nothing changed yet.
+- Also support version in setup.cfg. [ewjoachim]
+- Write version in ``setup.cfg``. [ewjoachim]
```
ewjoachim: hm, re-doing this.

ewjoachim: done

reinout:
Thanks! I just released 6.12.4, can you check if it all works for you?

ewjoachim:
Works great ! On both usual projects and setup.cfg projects. Thank you !

PR #232, opened 2017-08-15
Allow `{version}` in `tag-format`

### LINKED ISSUE #212: Tag formatter documentation
Not sure if this is common knowledge or not but apparently config parser uses interpolation ([issue28935](https://bugs.python.org/issue28935) and [issue20754](https://bugs.python.org/issue20754)) in python3. 
That means the `%` signs are causing the loading to crash.
It can be fixed by using `%%` instead but it did not occur to me when writing the documentation.
Do you thing this should be highlighted ?

leorochael:
Fixes #212

### DIFF
--- a/buildout.cfg
+++ b/buildout.cfg
@@ -53,7 +53,7 @@ zope.testrunner = 4.7.0
 [script]
 recipe = zc.recipe.egg
 dependent-scripts = true
-interpreter = python
+interpreter = py
 eggs =
     zest.releaser[recommended]
 
--- a/zest/releaser/pypi.py
+++ b/zest/releaser/pypi.py
@@ -423,30 +423,44 @@ class PypiConfig(object):
             return default
         return result
 
-    def tag_format(self):
-        """Return the string format for the tag name used in the release.
+    _tag_format_deprecated_message = "\n".join(line.strip() for line in """
+    `tag-format` contains deprecated `%%(version)s` format. Please change to:
+    
+    [zest.releaser]
+    tag-format = %s
+    """.strip().splitlines())
 
-        Override the default ``%(version)s`` in ~/.pypirc or setup.cfg using
-        a ``tag-format`` option::
+    def tag_format(self, version):
+        """Return the formatted tag that should be used in the release.
+
+        Configure it in ~/.pypirc or setup.cfg using a ``tag-format`` option::
 
             [zest.releaser]
-            tag-format = v%(version)s
+            tag-format = v{version}
 
-        Returns default of ``%(version)s`` when nothing has been configured.
+        ``tag-format`` must contain exaclty one formatting instruction: for the
+        ``version`` key.
+
+        Accepts also ``%(version)s`` format for backward compatibility.
+
+        The default format, when nothing has been configured, is ``{version}``.
         """
-        default = '%(version)s'
-        if self.config is None:
-            return default
-        try:
-            result = self.config.get(
-                'zest.releaser', 'tag-format')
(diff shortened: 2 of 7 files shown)


buildout.cfg
```diff
@@ -53,7 +53,7 @@ zope.testrunner = 4.7.0
 [script]
 recipe = zc.recipe.egg
 dependent-scripts = true
-interpreter = python
+interpreter = py
```
reinout: Why's this? 

(My guess: you use a virtualenv to get rid of the bootstrap.py? The change sounds OK then.)

zest/releaser/pypi.py
```diff
@@ -423,30 +423,31 @@ def version_levels(self):
...
 
-        Override the default ``%(version)s`` in ~/.pypirc or setup.cfg using
+        Override the default ``{version}`` in ~/.pypirc or setup.cfg using
         a ``tag-format`` option::
 
             [zest.releaser]
-            tag-format = v%(version)s
+            tag-format = v{version}
 
-        Returns default of ``%(version)s`` when nothing has been configured.
```
reinout: This line should stay, right?  (just change the % for { )

Your new line can also stay, but the original has its use, too.

zest/releaser/pypi.py
```diff
@@ -423,30 +423,31 @@ def version_levels(self):
             return default
         return result
 
-    def tag_format(self):
-        """Return the string format for the tag name used in the release.
+    def tag_format(self, version):
+        """Format the tag that should be used in the release.
```
reinout: "Return the formatted tag that ..."

zest/releaser/tests/pypi.txt
```diff
@@ -327,6 +327,50 @@ We try that again with this fixed up config file as input:
     tag_build =
     tag_svn_revision = false
 
+Formating release tags
```
reinout: Formating => formatting

reinout:
Looks like a perfect fix, thank you!

buildout.cfg
leorochael:
```
-interpreter = python
+interpreter = py
```

> Why's this?
> 
> (My guess: you use a virtualenv to get rid of the bootstrap.py? The change sounds OK then.)

I don't particularly want to get rid of `bootstrap.py` (I used it in fact), but I like using in-line virtualenvs, as in:

    virtualenv --no-setuptools --no-pip .

And running `bootstrap.py` after.

But as the commit message that introduced that change says, I'm changing this:

> So it doesn't conflict with an inline virtualenv.

Otherwise, running buildout causes `"Text file busy"` errors on (at least) Linux, when buildout tries to replace the Python executable that is running the buildout with a script.

zest/releaser/pypi.py
leorochael: >     -        Returns default of ``%(version)s`` when nothing has been configured.
> This line should stay, right? (just change the `%` for `{` )

I thought the first line of the doc was enough in establishing that `"{version}"` is the default format:

    Override the default ``{version}`` in ~/.pypirc or setup.cfg [...]

I'll reword to call out the default more explicitly.

leorochael:
@reinout, I've made the requested changes, and also added a deprecation message when the `%(version)s` format is used, so we can consider dropping it in the future.

reinout:
Thanks! I've released 9.12.3 with your patch. Can you double-check that it works?

leorochael:
Seems to work. I've tested it with `zc.recipe.egg` (though I couldn't actually upload the release as I'm not a PyPI developer for it.

PR #236, opened 2017-09-14
Add skip-missing-hooks option

leplatrem:
Hi there!

We have a use case where we really don't want the hooks to be skipped when they cannot be imported.

So I added an option, with current behaviour as default :)

I'm kinda puzzled with the way I should test this, I would be grateful if you had some insights about the best way to do it!

Thanks, and also thanks for this awesome tool ;)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,8 +4,8 @@ Changelog for zest.releaser
 6.12.5 (unreleased)
 -------------------
 
-- Add ``skip-missing-hooks`` option in ``setup.cfg`` to control whether the
-  release process should should fail when specified hooks cannot be imported.
+- Release process will now fail when specified hooks cannot be imported.
+  (`PR #236 <https://github.com/zestsoftware/zest.releaser/pulls/236>`_)
 
 
 6.12.4 (2017-08-30)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,8 @@ Changelog for zest.releaser
 6.12.5 (unreleased)
 -------------------
 
-- Nothing changed yet.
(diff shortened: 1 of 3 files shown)


Natim:
`TypeError: getboolean() got an unexpected keyword argument 'fallback'` with Python2.7

Natim:
I don't see how the failure in the pypy env are related to that fix.

reinout:
The code itself looks fine. But.... I always dislike extra options on principle :-)

In this case, my thought is "shouldn't this be the default behaviour"? Why would you specify a hook in your `setup.cfg` if you didn't intend to run it? Perhaps because not all developers have installed the necessary library: that's the only reason I can think of right now.

Am I missing something?

My current thoughts:

- Simply fail when a hook is missing. (So: no need for your extra option).

- Or use your extra option, but to make "fail" the *default*. In this case, you *do allow* for missing hooks, but you have to do the extra config work if you're in this corner case. My assumption is that your build tools normally make sure all (dev) dependencies are properly installed, so the hook should really be there.

What are your thoughts?

reinout:
I've re-run the pypy job on travis: it passes now. It already looked like a temporary glitch. We're testing lots of python versions, so small glitches show up more often :-)

leplatrem:
Thanks @reinout for your feedback!

Indeed, I would also be in favor of failing when hook missing! But my intention was to keep the current behavior as default, hence the option.

I will change that immediately then :)

leplatrem:
I updated the code (pypy job might have to be restarted)

reinout:
I've released 6.12.5 with your fix!

Natim:
\o/ thanks

PR #240, opened 2017-09-25
Sort filenames to be uploaded

reinout:


reinout:
@mauritsvanrees: test failures are only on the pypy tests. Somehow one of the post-release calls to either bzr or git (I've seen both) fail. So that looks fine.

I have no idea why the pypy one fails, though.

reinout:
@mauritsvanrees: I'm assuming a `.sort()` is small enough to release it right away :-)

reinout:
Released as 6.12.5

PR #246, opened 2017-11-10
Use list of arguments for execute_command, not a string.

### LINKED ISSUE #233: Use list of arguments for execute_command, not a string
In #228 the code was changed to quote the path of the git-repos. This is only necessary since the code uses strings as commands, instead of a list of command arguments.

**This is bad practice!** and may be insecure. Just imaging (in this use case) the path contains a quote-character `"`. Then the command will fail or even damage things.

**Best-practice** is to pass a list of arguments to Popen (and siblings) This avoids *all* quoting issues, since there is no command string which needs to be parsed to split it into arguments.

This may not be an issue in the case of cloning the repo, but it is an issue in general cases.

htgoebel:
This avoids wrong quoting in arguments if one e.g. passes a
double-quote in a message.

- Convert all command-strings into lists.
- New function ``utils.excute_commands()`` (plural).
- New function ``utils.format_commands()``.
- Incompatible: ``git.cmd_checkout_from_tag()`` now returns a list
  of commands, not a single one.
- Adjusted tests.

Closes #233.

### DIFF
--- a/zest/releaser/git.py
+++ b/zest/releaser/git.py
@@ -38,7 +38,7 @@ class Git(BaseVersionControl):
         return dir_name
 
     def available_tags(self):
-        tag_info = execute_command('git tag')
+        tag_info = execute_command(['git', 'tag'])
         tags = [line for line in tag_info.split('\n') if line]
         logger.debug("Available tags: %r", tags)
         return tags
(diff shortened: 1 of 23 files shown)


htgoebel:
Tests pass, ready to merge.

Thanks to the extensive test-suite, it made working out bugs a breeze.

zest/releaser/git.py
```diff
@@ -64,25 +64,25 @@ def tag_url(self, version):
...
-        return "git log %s..HEAD" % version
+        return ['git', 'log', '%s..HEAD' % version]
 
     def cmd_create_tag(self, version, sign=False):
         msg = "Tagging %s" % (version,)
-        cmd = 'git tag %s -m "%s"' % (version, msg)
+        cmd = ['git', 'tag', version, '-m', msg]
         if sign:
-            cmd += " --sign"
+            cmd.append["--sign"]
```
reinout: I only noticed this one when trying to release it today:

      File "/Users/reinout/local/zest.releaser/zest/releaser/git.py", line 84, in cmd_create_tag
        cmd.append["--sign"]
    TypeError: 'builtin_function_or_method' object has no attribute '__getitem__'

:-) I'll fix it so that I can release zest.releaser again :-)

htgoebel:
Huh! What a stupid bug! Looks like this is not covered by a test-case.

reinout:
Yes. That one was missing from the test case. It was easy to add, though. Fixed in the mean time.

PR #255, opened 2017-11-24
Try output encodings in the right order please

### LINKED ISSUE #254: prerelease mangles UTF-8 in the changelog
Steps to reproduce:

- git clone https://github.com/mgedmin/objgraph && cd objgraph
- prerelease
- git diff

Expected result

- release date is changed, nothing else is

Actual result

- "Kristján" is mangled into "KristjÃ¡n"

mgedmin:
Set iteration order is not defined in Python.  This loop is one where the order *matters*, otherwise we could be "upgrading" to UTF-8 without any reason, leading to bugs like #254.

I have not tested that this fix works.

I don't currently have the time to write a unit test that would expose the bug.

### DIFF
--- a/zest/releaser/utils.py
+++ b/zest/releaser/utils.py
@@ -71,7 +71,7 @@ def write_text_file(filename, contents, encoding=None):
         # We might have added something to the contents (a changelog entry)
         # that does not fit the detected encoding.  So we try a few encodings.
         orig_encoding = encoding
-        encodings = set([orig_encoding, OUTPUT_ENCODING, 'utf-8'])
+        encodings = [orig_encoding, OUTPUT_ENCODING, 'utf-8']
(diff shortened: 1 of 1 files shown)


mgedmin:
Also, the insidious thing about this bug is that `write_text_file()` first reports

```
            logger.debug("Writing to %s with its original encoding %s",
                         filename, encoding)
```

and *then* it silently upgrades that encoding to something else, without informing the user.  You may want to fix that too!

reinout:
Ouch! Good catch. That `set()` is a small optimization that actually breaks the functionality!

95e3bd3daa1617e85aee56ba93c961be8784e7da fixes the logging.

reinout:
Released in 6.13.2

mgedmin:
Thank you!

PR #257, opened 2017-12-14
Write text files with correct encoding on Python3

### LINKED ISSUE #256: Crash when updating changelog due to incorrect encoding
When running `prerelease`, I'm getting the same error as in #210, except the incorrect encoding comes from zest.releaser itself. I'm running it on Python 3.6.0 on a Windows machine, and the changelog is encoded as UTF-8. After the crash, the changes are left on the file, but encoded as `cp1252`.

The problem seems to be that zest.releaser simply ignores the encoding when writing text files on Python 3, which means the platform uses the default system encoding (obtained using `locale`).

andreparames:
This patch seems to fix #256, but I've only tested on my Windows machine.

### DIFF
--- a/zest/releaser/utils.py
+++ b/zest/releaser/utils.py
@@ -59,31 +59,35 @@ def loglevel():
 
 
 def write_text_file(filename, contents, encoding=None):
-    if six.PY2 and isinstance(contents, six.text_type):
-        # Python 2 unicode needs to be encoded.
-        if encoding is None:
-            encoding = OUTPUT_ENCODING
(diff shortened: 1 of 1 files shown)


andreparames:
The build on Python 2 failed, but doesn't seem to be related to my change.

andreparames:
@mauritsvanrees thanks, I've rebased and Travis passed :tada:

reinout:
I've added a changelog entry and released it as 6.13.3: can you double-check if it works for you now?

Issue #262, opened 2018-02-13
postrelease fails with AttributeError: 'NoneType' object has no attribute 'find' in empty git repository


mgedmin:
I accidentally ran `postrelease` in the wrong directory and got this error:

```
Traceback (most recent call last):
  File "/home/mg/.venv/bin/postrelease", line 11, in <module>
    sys.exit(main())
  File "/home/mg/.venv/local/lib/python2.7/site-packages/zest/releaser/postrelease.py", line 102, in main
    postreleaser.run()
  File "/home/mg/.venv/local/lib/python2.7/site-packages/zest/releaser/baserelease.py", line 351, in run
    self.prepare()
  File "/home/mg/.venv/local/lib/python2.7/site-packages/zest/releaser/postrelease.py", line 50, in prepare
    self._ask_for_new_dev_version()
  File "/home/mg/.venv/local/lib/python2.7/site-packages/zest/releaser/postrelease.py", line 65, in _ask_for_new_dev_version
    current = utils.cleanup_version(current)
  File "/home/mg/.venv/local/lib/python2.7/site-packages/zest/releaser/utils.py", line 174, in cleanup_version
    if version.find(w) != -1:
AttributeError: 'NoneType' object has no attribute 'find'
```

Steps to reproduce:

- `git init /tmp/empty`
- `cd /tmp/empty`
- `postrelease`

mgedmin:
(Edited the original description to accurately describe what the problem is.  The original email notification was quite misleading, because I got confused.  Sorry about that!)

mauritsvanrees:
In `prerelease` we test for this and exit with a critical error. We should do the same here in postrelease. I will make a PR shortly.

mauritsvanrees:
Thanks for the report. Fixed in 6.13.5.

mgedmin:
Thank you!

PR #263, opened 2018-02-16
Quit in postrelease when we cannot find a version.

mauritsvanrees:
Fixes issue #262.

There were several checks like this already. I made them the same: do not compare with `None`, but check for a False value. That would catch a found but empty version as well.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,9 @@ Changelog for zest.releaser
 6.13.5 (unreleased)
 -------------------
 
-- Nothing changed yet.
+- Quit in ``postrelease`` when we cannot find a version.
+  Fixes `issue #262 <https://github.com/zestsoftware/zest.releaser/issues/262>`.
+  [maurits]
 
 
 6.13.4 (2018-02-05)
--- a/zest/releaser/bumpversion.py
+++ b/zest/releaser/bumpversion.py
@@ -76,7 +76,7 @@ class BumpVersion(baserelease.Basereleaser):
         """
         original_version = self.vcs.version
         logger.debug("Extracted version: %s", original_version)
-        if original_version is None:
+        if not original_version:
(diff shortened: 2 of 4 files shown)


mauritsvanrees:
Travis passes, but there is an allowed error for PyPy. It passed on master. It seems to be caused by my change:

```
File "/home/travis/build/zestsoftware/zest.releaser/zest/releaser/tests/functional-with-hooks.txt", line 121, in functional-with-hooks.txt
Failed example:
    postrelease.main()
Exception raised:
      File "/opt/python/pypy2.7-5.8.0/lib-python/2.7/doctest.py", line 1315, in __run
        compileflags, 1) in test.globs
      File "<doctest functional-with-hooks.txt[19]>", line 1, in <module>
        postrelease.main()
      File "/home/travis/build/zestsoftware/zest.releaser/zest/releaser/postrelease.py", line 106, in main
        postreleaser.run()
      File "/home/travis/build/zestsoftware/zest.releaser/zest/releaser/baserelease.py", line 351, in run
        self.prepare()
      File "/home/travis/build/zestsoftware/zest.releaser/zest/releaser/postrelease.py", line 49, in prepare
        sys.exit(1)
      File "/home/travis/build/zestsoftware/zest.releaser/zest/releaser/tests/functional.py", line 36, in _exit
        raise RuntimeError(msg)
    RuntimeError: SYSTEM EXIT (code=1)
```

mauritsvanrees:
Okay, previously PyPy passed on master, but that seems just luck. There are earlier runs that [fail in the same way](https://travis-ci.org/zestsoftware/zest.releaser/jobs/318673866). And I have restarted the PyPy job and I get a similar error but at a different spot. So the PyPy job is indeed unstable.
Would be nice to get that fixed, but that is not for this PR. I will merge.

PR #273, opened 2018-03-23
Pin pkginfo to 1.4.2.

mauritsvanrees:
Latest twine needs `pkginfo>=1.4.2`.
A few more packages are pulled in.
Updated zc.buildout to latest.

This lets buildout finish.
I did the same on PR #272 but one test fails there, and I wonder if that is because of the other changes in that PR.


reinout:
LGTM

Issue #271, opened 2018-03-16
% in password breaks upload


brainwane:
My password was:

`cfEc9!vIqv%bjv4Imy^1@WJali0r6w0%GM%HwpJlzbm*0`

From a `fullrelease` error log:

~~~
Showing last few lines...
Writing mauritstestpackage-sumana-version-1.23/setup.cfg
creating dist
Creating tar archive
removing 'mauritstestpackage-sumana-version-1.23' (and everything under it)

Upload to testpypi (Y/n)? Y
Traceback (most recent call last):
  File "/home/sumanah/.virtualenvs/twine-test-15-mar-2018/bin/fullrelease", line 11, in <module>
    sys.exit(main())
  File "/home/sumanah/.virtualenvs/twine-test-15-mar-2018/lib/python3.6/site-packages/zest/releaser/fullrelease.py", line 27, in main
    releaser.run()
  File "/home/sumanah/.virtualenvs/twine-test-15-mar-2018/lib/python3.6/site-packages/zest/releaser/baserelease.py", line 353, in run
    self.execute()
  File "/home/sumanah/.virtualenvs/twine-test-15-mar-2018/lib/python3.6/site-packages/zest/releaser/release.py", line 75, in execute
    self._release()
  File "/home/sumanah/.virtualenvs/twine-test-15-mar-2018/lib/python3.6/site-packages/zest/releaser/release.py", line 324, in _release
    self._upload_distributions(package)
  File "/home/sumanah/.virtualenvs/twine-test-15-mar-2018/lib/python3.6/site-packages/zest/releaser/release.py", line 174, in _upload_distributions
    self._retry_twine('upload', server, filename)
  File "/home/sumanah/.virtualenvs/twine-test-15-mar-2018/lib/python3.6/site-packages/zest/releaser/release.py", line 191, in _retry_twine
    repository = self._get_repository(server)
  File "/home/sumanah/.virtualenvs/twine-test-15-mar-2018/lib/python3.6/site-packages/zest/releaser/release.py", line 180, in _get_repository
    **self.pypiconfig.get_server_config(server))
  File "/home/sumanah/.virtualenvs/twine-test-15-mar-2018/lib/python3.6/site-packages/zest/releaser/pypi.py", line 186, in get_server_config
    password = self.config.get(server, 'password')
  File "/usr/local/lib/python3.6/configparser.py", line 800, in get
    d)
  File "/usr/local/lib/python3.6/configparser.py", line 394, in before_get
[...]

PR #274, opened 2018-03-24
Get PyPI password raw, without interpolation.

mauritsvanrees:
If you had a password with a percentage sign, you could get an error.
Fixes https://github.com/zestsoftware/zest.releaser/issues/271.

There might be other places where we need to do this, but it is hard to say where interpolation may be useful and where it may be harmful.


reinout:
LGTM

Issue #264, opened 2018-02-21
Charset confusion in prerelease


mgedmin:
I was cutting the 2.17.0 release of [irclog2html](https://github.com/mgedmin/irclog2html), and saw this:

```
$ prerelease
...
Changelog entries for version 2.17.0:

2.17.0 (unreleased)
-------------------

- Support `ii <https://tools.suckless.org/ii/>`_ logs which use Unix
  timestamps (https://github.com/mgedmin/irclog2html/pull/21).
  Pull request by CÃ©dric Krier.
Enter version [2.17.0]: 
```

Note how the changelog text was misidentified as Latin-1 and displayed wrong.  The file itself is UTF-8:

```
$ head -n 10 CHANGES.rst 
Changelog
=========

2.17.0 (2018-02-21)
-------------------

- Support `ii <https://tools.suckless.org/ii/>`_ logs which use Unix
  timestamps (https://github.com/mgedmin/irclog2html/pull/21).
  Pull request by Cédric Krier.
```

Luckily the text is only displayed wrong, not actually written wrong into the file.

My locale settings are correct (`locale charmap` prints `UTF-8`).  I'm using Python 2.  `prerelease --version` prints `prerelease: error: unrecognized arguments: --version`, but anyway I just did a `pip install -U zest.releaser` so it's the latest.

mauritsvanrees:
To detect the encoding, on Python 2 we use `chardet` and on Python 3 we use `tokenize`. See https://github.com/zestsoftware/zest.releaser/blob/6.13.5/zest/releaser/utils.py#L93-L108

In debug mode, I see this with Python 2:

```
$ prerelease -v
...
DEBUG: Checking CHANGES.rst
DEBUG: EUC-TW Taiwan prober hit error at byte 213
DEBUG: utf-8  confidence = 0.505
DEBUG: SHIFT_JIS Japanese confidence = 0.01
DEBUG: EUC-JP Japanese confidence = 0.01
DEBUG: GB2312 Chinese confidence = 0.01
DEBUG: EUC-KR Korean confidence = 0.01
DEBUG: CP949 Korean confidence = 0.01
DEBUG: Big5 Chinese confidence = 0.01
DEBUG: EUC-TW not active
DEBUG: windows-1251 Russian confidence = 0.01
DEBUG: KOI8-R Russian confidence = 0.01
DEBUG: ISO-8859-5 Russian confidence = 0.01
[...]

PR #275, opened 2018-03-26
Encoding config option

mauritsvanrees:
Fixes issue #264, see discussion there.

I fought a bit with the difference between a few possibilities to read this:
- create a `SetupConfig` object and read the encoding option directly in the calling code (`if setup_config.config.has_option(...)`
- create a `PypiConfig` object, which also reads `setup.cfg` and use a new method `encoding` on that class.

I opted for the second way.
That also meant that I found an undocumented option `history_file` and changed that to use the second way. So I included that in this PR. I kept the commits separate, so you can view them separately if you want.

### DIFF
--- a/zest/releaser/pypi.py
+++ b/zest/releaser/pypi.py
@@ -304,6 +304,31 @@ class PypiConfig(BaseConfig):
             return default
         return result
 
+    def encoding(self):
+        """Return encoding to use for text files.
+
+        Mostly the changelog and if needed `setup.py`.
+
+        The encoding cannot always be determined correctly.
+        This setting is a way to fix that.
+        See https://github.com/zestsoftware/zest.releaser/issues/264
+
+        Configure this by adding an ``encoding`` option, either in the
+        package you want to release, or in your ~/.pypirc::
+
+            [zest.releaser]
+            encoding = utf-8
+        """
+        default = ''
+        if self.config is None:
+            return default
+        try:
+            result = self._get_text(
+                'zest.releaser', 'encoding', default=default, raw=True)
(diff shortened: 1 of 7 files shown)


zest/releaser/pypi.py
```diff
@@ -304,6 +304,62 @@ def extra_message(self):
             return default
         return result
 
+    def history_file(self):
```
reinout: In combination with `self._get_text()` I originally thought "we're grabbing the textual content of the history file".
The cause for the misunderstanding is that I'm only seeing the changes and not the whole :-)

But... perhaps `history_file_path` might be a tad clearer. Depends on the other method names, of course. I should read the whole source code again some day as I don't know it from the top of my head :-)

mauritsvanrees: The code was already looking in the config for `history_file`, so I say it is best that we keep it that way.

mauritsvanrees:
> One extra comment: read_text_file() has a fallback_encoding argument. Is that necessary? Or could the function look it up itself?
> (I assume "it is necessary", though).

Technically, we could call `config = pypi.PypiConfig(); fallback_encoding = config.encoding()` in `read_text_file`. But then you instantiate the PypiConfig multiple times, which sounds wrong. It doesn't take long, so it may be okay. But I am also half wondering if `utils.py` would be able to find the right `setup.cfg` if we are not in the root directory. Maybe best to leave it this way.

I guess there is room for a more thorough overhaul of the package if we want to clean it up, rethinking decisions that we made in the past, and making it more clear what other packages (especially zest.releaser hooks) can depend on as api.

reinout:
Ok, clear!

PR #276, opened 2018-03-26
Fixed more ugly printing.

mauritsvanrees:
Printing the twine command for retrying would give this:

    Command failed: "t w i n e ' ' u p l o a d"

And with Python2.7, in this and various other spots, a ``u'unicode'`` marker would be visible. This is because several things were logged with ``%r``.

In most cases, the only reason for the raw format was so that it was clear where the command started and ended, because it would be enclosed in 'quotes'. But for unicode this becomes u'quotes', which is ugly. Note that we use `from __future__ import unicode_literals` so all strings in the code are actually unicode.

So instead of `%r` we now quote it ourselves with `'%s'`.

### DIFF
--- a/zest/releaser/git.py
+++ b/zest/releaser/git.py
@@ -40,7 +40,7 @@ class Git(BaseVersionControl):
     def available_tags(self):
         tag_info = execute_command(['git', 'tag'])
         tags = [line for line in tag_info.split('\n') if line]
-        logger.debug("Available tags: %r", tags)
+        logger.debug("Available tags: '%s'", ' '.join(tags))
         return tags
 
     def prepare_checkout_dir(self, prefix):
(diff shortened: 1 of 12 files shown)


zest/releaser/git.py
```diff
@@ -40,7 +40,7 @@ def name(self):
     def available_tags(self):
         tag_info = execute_command(['git', 'tag'])
         tags = [line for line in tag_info.split('\n') if line]
-        logger.debug("Available tags: %r", tags)
+        logger.debug("Available tags: '%s'", ' '.join(tags))
```
reinout: Join with a comma+space?
OTOH, git prints them without...

mauritsvanrees: Ah, true, a comma would be clearer.
I'll change and locally and merge manually.

mauritsvanrees:
Merged manually.

PR #279, opened 2018-04-18
Towncrier support [do not merge]

mauritsvanrees:
For some Plone packages we want to start using [`towncrier`](https://pypi.org/project/towncrier/) to handle updating `CHANGES.rst`, just like for example [`pip` does](https://pip.pypa.io/en/latest/development/#adding-a-news-entry). For explanation, see https://github.com/plone/plone.releaser/issues/17. (`plone.releaser` is a wrapper around `zest.releaser` that we use when creating releases for Plone packages.)

This may need to be done in a separate small plugin. But it needs a few changes in `zest.releaser`. Mainly: when `towncrier` updates the `CHANGES.rst`, `zest.releaser` should not touch that file at all.

So I made two commits:
- one with a few minor changes in `zest.releaser`: 39b465ed642b7420b0bfe5e0d79b567978b5b605
- one with the hooks for calling `towncrier`: acd9b97a02fb681c0a0d72aaa253a52516aa8a90

The second part may need to be moved to a separate plugin, and I am happy to do so. It *could* be okay to include it in `zest.releaser`, and it at least makes it easier to try out this PR.

I have used this code to make release [4.0.1](https://github.com/collective/collective.recipe.backup/tree/4.0.1) of [`collective.recipe.backup`](https://github.com/collective/collective.recipe.backup). Most important are these commits:
- [prerelease](https://github.com/collective/collective.recipe.backup/commit/935c48629b4b71d1b98d573d7adbaaf6825d01b9)
- [postrelease](https://github.com/collective/collective.recipe.backup/commit/3482b0ed601e8916a51ffd623458374821576e0a)

So:
- review welcome on the first commit with the minor zest.releaser changes
- feedback welcome on the towncrier hooks and whether that should be a separate plugin

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -110,5 +110,5 @@ setup(name='zest.releaser',
               'zest.releaser.preparedocs:prepare_entrypoint_documentation',
               ],
 
-          },
+      },
       )
--- a/setup.py
+++ b/setup.py
@@ -67,6 +67,8 @@ setup(name='zest.releaser',
               'check-manifest',
               'pyroma',
               'readme_renderer',
+              'toml',
+              'towncrier',
               'wheel',
               ],
           'test': [
@@ -91,6 +93,8 @@ setup(name='zest.releaser',
           # our entry point implementation.
           'zest.releaser.prereleaser.middle': [
               'datacheck = zest.releaser.prerelease:datacheck',
+              # Call towncrier to update the changelog with newsfragments:
+              'call_towncrier = zest.releaser.towncrier_hooks:call_towncrier',
               ],
           'zest.releaser.releaser.middle': [
               'datacheck = zest.releaser.release:datacheck',
@@ -108,6 +112,15 @@ setup(name='zest.releaser',
           'zest.releaser.prereleaser.before': [
               'preparedocs = ' +
               'zest.releaser.preparedocs:prepare_entrypoint_documentation',
+              # Check if towncrier is available and configured:
+              'check_towncrier = zest.releaser.towncrier_hooks:check_towncrier',
+              ],
+          'zest.releaser.postreleaser.before': [
+              # Towncrier for news.
+              # We only need to check if towncrier is available and
+              # configured for this project.
+              # Then the history should not be changed.
+              'check_towncrier = zest.releaser.towncrier_hooks:check_towncrier',
               ],
 
       },
--- a/zest/releaser/postrelease.py
+++ b/zest/releaser/postrelease.py
@@ -22,6 +22,8 @@ DATA.update({
     'dev_version_template': 'Template for development version number',
     'development_marker': 'String to be appended to version after postrelease',
     'new_version': 'New version, without development marker (so 1.1)',
+    'update_history': 'Should zest.releaser update the history file?',
+    'update_version': 'Should zest.releaser update the version file?',
 })
 
 
@@ -40,6 +42,8 @@ class Postreleaser(baserelease.Basereleaser):
             dev_version_template=DEV_VERSION_TEMPLATE,
             development_marker=self.pypiconfig.development_marker(),
             history_header=HISTORY_HEADER,
+            update_history=True,
+            update_version=True,
         ))
 
     def prepare(self):
@@ -52,9 +56,12 @@ class Postreleaser(baserelease.Basereleaser):
 
     def execute(self):
         """Make the changes and offer a commit"""
-        self._write_version()
-        self._change_header(add=True)
-        self._write_history()
+        if self.data['update_version']:
+            self._write_version()
+        if self.data['update_history']:
+            self._change_header(add=True)
+            self._write_history()
+        # Always ask to commit, because an entrypoint may have made changes.
         self._diff_and_commit()
         self._push()
 
--- a/zest/releaser/towncrier_hooks.py
+++ b/zest/releaser/towncrier_hooks.py
@@ -1,0 +1,99 @@
+# -*- coding: utf-8 -*-
+from zest.releaser import utils
+import logging
+import os
+import pkg_resources
+import toml
+
+# towncrier might become a conditional dependency of zest.releaser,
+# for example in the recommended list.
+# If we split this into a separate package, it should be a hard dependency.
+try:
+    pkg_resources.get_distribution('towncrier')
+except pkg_resources.DistributionNotFound:
+    HAS_TOWNCRIER = False
+else:
+    HAS_TOWNCRIER = True
+
+
+logger = logging.getLogger('towncrier_hooks')
+TOWNCRIER_MARKER = '_towncrier_applicable'
+TOWNCRIER_CONFIG_FILE = 'pyproject.toml'
+
+
+def load_config():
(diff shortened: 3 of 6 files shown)


zest/releaser/towncrier_hooks.py
```diff
@@ -0,0 +1,99 @@
...
+    HAS_TOWNCRIER = True
+
+
+logger = logging.getLogger('towncrier_hooks')
+TOWNCRIER_MARKER = '_towncrier_applicable'
+TOWNCRIER_CONFIG_FILE = 'pyproject.toml'
+
+
+def load_config():
+    fn = os.path.join(TOWNCRIER_CONFIG_FILE)
```
reinout: What does "fn" mean? It sounds like "function". Better have a more descriptive name.

zest/releaser/towncrier_hooks.py
```diff
@@ -0,0 +1,99 @@
...
+    if not os.path.exists(fn):
+        return
+    with open(fn, 'r') as conffile:
+        config = toml.load(conffile)
+    if 'tool' not in config:
+        return
+    config = config['tool']
```
reinout: I don't really like this re-write of an existing variable name. That will lead to errors/misunderstandings.

zest/releaser/towncrier_hooks.py
```diff
@@ -0,0 +1,99 @@
...
+    if 'towncrier' not in config:
+        return
+    config = config['towncrier']
```
reinout: Another one. Is this common practice for .toml files?

zest/releaser/towncrier_hooks.py
```diff
@@ -0,0 +1,99 @@
...
+def _check_towncrier():
+    # First check if the towncrier tool is available.
+    # towncrier might be on the PATH but not importable or the other way around.
+    # So could be tricky to know for sure that it is available.
+    # Maybe call 'towncrier --help' and see if that gives an error.
+    if not HAS_TOWNCRIER:
+        return False
+    # Read pyproject.toml with the toml package.
+    if not load_config():
+        return False
```
reinout: So.... if towncrier is importable (from some other project or so), you still have to do some config yourself?
That's good, as otherwise you could get it everywhere when we include it in zest.releaser.

zest/releaser/towncrier_hooks.py
```diff
@@ -0,0 +1,99 @@
...
+    return True
+
+
+def check_towncrier(data):
```
reinout: The names `_check_towncrier()` and `check_towncrier()` are very similar. The underscore version should perhaps be named something like `_is_towncrier_enabled()`?

zest/releaser/towncrier_hooks.py
```diff
@@ -0,0 +1,99 @@
...
+        return False
+    # Check if towncrier should be applied.
+    result = _check_towncrier()
+    if result:
+        logger.debug('towncrier should be applied.')
+        # zest.releaser should not update the history.
+        # towncrier will do that.
+        data['update_history'] = False
+    else:
+        logger.debug('towncrier cannot be applied.')
```
reinout: "cannot be applied"? I could also be "is not available", right?

zest/releaser/towncrier_hooks.py
```diff
@@ -0,0 +1,99 @@
...
+    data[TOWNCRIER_MARKER] = result
+    return result
+
+
+def call_towncrier(data):
+    """Entrypoint: run towncrier when applicable."""
```
reinout: ".... when ENABLED AND AVAILABLE"?

I get the feeling that this will all be clearer as a separate plugin.

zest/releaser/towncrier_hooks.py
```diff
@@ -0,0 +1,99 @@
...
+    if not check_towncrier(data):
+        return
+    cmd = ['towncrier', '--version', data['new_version'], '--yes']
```
reinout: This might not always work. When you install zest.releaser in a buildout, you'll call `bin/fullrelease`. A simple `towncrier` call will then fail as it should be `bin/towncrier`.

setup.py
```diff
@@ -67,6 +67,8 @@ def read(filename):
               'check-manifest',
               'pyroma',
               'readme_renderer',
+              'toml',
```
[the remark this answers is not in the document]
mauritsvanrees: It kind of sounds familiar, yes. I don't see an issue or PR that seems related though.

zest/releaser/postrelease.py
```diff
@@ -22,6 +22,8 @@
     'dev_version_template': 'Template for development version number',
     'development_marker': 'String to be appended to version after postrelease',
     'new_version': 'New version, without development marker (so 1.1)',
+    'update_history': 'Should zest.releaser update the history file?',
+    'update_version': 'Should zest.releaser update the version file?',
```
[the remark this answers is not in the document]
mauritsvanrees: We would still be asking the user for a version, with a suggestion from us, and this would end up in the data, ready for future plugins to use it. In the towncrier command I use that version from the data as well.
But I will remove this `update_version` part. Let's keep it restricted to just the `update_history` part that I really need.

zest/releaser/towncrier_hooks.py
mauritsvanrees: This part was copied from the `towncrier` code, and only slightly adapted. I can indeed improve it.

zest/releaser/towncrier_hooks.py
mauritsvanrees: Same.

zest/releaser/towncrier_hooks.py
mauritsvanrees: Maybe. It is in towncrier.
It can be okay: you start with the complete config and drill down until you have the part of the config that you really need, and don't care about the abandoned parts.

zest/releaser/towncrier_hooks.py
mauritsvanrees: Correct. You need a `pyproject.toml` properly configured from `towncrier`, otherwise we do nothing, because calling `towncrier` would fail.

zest/releaser/towncrier_hooks.py
mauritsvanrees: That clarification can work yes.

zest/releaser/towncrier_hooks.py
mauritsvanrees: I might want to try a few spots, yes, for example look for a `towncrier` file in the same directory as the `fullrelease` file.
It does not look like there is an api we can rely on.

mauritsvanrees:
Thanks for the feedback. I will create a plugin for this, probably in the Plone collective.
I will close this PR and create a new one for clarity, where I will only do the `update_history` stuff, and not the `update_version`.

zest/releaser/towncrier_hooks.py
mauritsvanrees: Nice: you can try to find an executable on a path by using `distutils.spawn.find_executable`. See https://stackoverflow.com/a/12611523/621201

PR #280, opened 2018-04-24
Added key update_history in data.

mauritsvanrees:
This replaces PR #279.

This is in prerelease and postrelease data.
Plugins can use this to tell `zest.releaser` (and other plugins) to not touch the history, presumably because the plugin itself handles it.

Default value is True: do update history.

The `update_history` key will be used by a towncrier plugin that will update the changelog with news fragments.

The towncrier plugin will add an internal data key `_towncrier_applicable`.
With this, you would get this warning when running zest.releaser:

    Internal detail: key(s) ['_towncrier_applicable'] are not documented

So I have changed this internal check to allow undocumented data keys when they start with underscore.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,11 @@ Changelog for zest.releaser
 6.14.1 (unreleased)
 -------------------
 
+- Added key ``update_history`` in prerelease and postrelease data.
+  Plugins can use this to tell ``zest.releaser`` (and other plugins)
+  to not touch the history, presumably because the plugin handles it.
+  [maurits]
+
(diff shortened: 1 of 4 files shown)


mauritsvanrees:
I have create a plugin here: https://github.com/collective/zestreleaser.towncrier

reinout:
I shouldn't move jobs... that's bad for my review queue :-)

reinout:
I assume you want to release it? Feel free. I'm not doing it yet as you might have other things thought out while I was to lazy to review this :-)

mauritsvanrees:
Released in zest.releaser 6.15.0.

Issue #238, opened 2017-09-21
Need a option to sign when uploading (pass `--sign` to `twine upload`)


htgoebel:
For PyInstaller, we are using a [`releaser.before_upload` hook](https://github.com/pyinstaller/pyinstaller/blob/develop/setup.cfg#L9) to [sign releases](https://github.com/pyinstaller/pyinstaller/blob/develop/PyInstaller/utils/release.py#L27).
Now twine no longer allows uploading the additional file but fail (see below) with 
```
  Unknown distribution format: 'PyInstaller-3.3.tar.gz.asc'
```
##### Proposed solution
`twine upload` has an option `sign`, so all we need is a way to pass this option to `twine upload`.

##### Traceback
```
$ release
Tag needed to proceed, you can use the following command:
git tag 3.3 -m "Tagging 3.3"
Run this command (Y/n)? 
Check out the tag (for tweaks or pypi/distutils server upload) (Y/n)? 
INFO: Doing a checkout...
…
Writing PyInstaller-3.3/setup.cfg
creating dist
Creating tar archive
removing 'PyInstaller-3.3' (and everything under it)

Signing file /tmp/PyInstaller-3.3-8mqhmxfv/gitclone/dist/PyInstaller-3.3.tar.gz

… [Enter passphrese for gpgp key]
Upload to testpypi (Y/n)?     
Traceback (most recent call last):
  File "/tmp/py-web/bin/release", line 11, in <module>
    sys.exit(main())
  File "/tmp/py-web/lib64/python3.5/site-packages/zest/releaser/release.py", line 332, in main
    releaser.run()
  File "/tmp/py-web/lib64/python3.5/site-packages/zest/releaser/baserelease.py", line 353, in run
    self.execute()
  File "/tmp/py-web/lib64/python3.5/site-packages/zest/releaser/release.py", line 71, in execute
    self._release()
  File "/tmp/py-web/lib64/python3.5/site-packages/zest/releaser/release.py", line 317, in _release
    self._upload_distributions(package)
  File "/tmp/py-web/lib64/python3.5/site-packages/zest/releaser/release.py", line 167, in _upload_distributions
    self._retry_twine('upload', server, filename)
  File "/tmp/py-web/lib64/python3.5/site-packages/zest/releaser/release.py", line 185, in _retry_twine
    package_file = PackageFile.from_filename(filename, comment=None)
[...]

PR #284, opened 2018-05-24
Use twine.cli.dispatch() to run twine

htgoebel:
See https://github.com/zestsoftware/zest.releaser/issues/238#issuecomment-353128610:

I tried this and found myself re-implementing quite some parts of `twine.commands.upload#upload()`. Not only one would need to filter out the `.asc` files first, but also `add_gpg_signature` to each file to be uploaded, if the signature exists. Although this would still be only a view lines, keeping up with twine features and PyPI API changes would become a cat-and-mice-game in he long run.

I asked for another high-level api (see https://github.com/pypa/twine/issues/194#issuecomment-353126091), one of which `twine.cli.dispatch()` could be. This would allow something like:
```
dispatch(twine_command, '-r', server, *files_in_dist)
```


htgoebel:
I know this code has conflict, but I wanted to have a pull-request to refer to in the pypa/twine pull/361 for the case I'm deleting my fork.

htgoebel:
I just rebased on current master and resolved the conflicts.

The discussion about an API for twine did not show any results and what is discussed there is much too detailed. Thus I suggest to not wait for twine providing some documented API, but use this high-level function call.

Please note: The CI-tests for Python 2.7 fail when trying to install git-svn, see#307. Otehr tests fail with a stale build.

reinout:
Let's hope that twine doesn't remove/hide the perfectly fine cli "api".
I just totally don't get why twine seems to dislike its own defaults. If you use it from the outside, the defaults should be banned and you should do all sorts of config work, replicating what twine itself already does.

A well, we can always just call it on the command line.... That's what we did with `setup.py upload`.

zest/releaser/tests/pypi.txt
```diff
@@ -43,22 +43,6 @@ There are two styles of ``.pypirc`` files.  The old one just for pypi:
     >>> pypiconfig.distutils_servers()
     ['pypi']
 
-With ``get_server_config`` we get the settings that we pass to ``twine``:
-
-    >>> from pprint import pprint
```
reinout: This breaks some tests later on.

htgoebel: That's curious. Tests pass locally.  Anyway I see you fixed this in e7742a67f1d6

PR #285, opened 2018-06-01
Use Annotated Tags in Git

kojiromike:
According to [git's own documentation](https://git-scm.com/docs/git-tag):

> Annotated tags are meant for release while lightweight tags are meant for private or temporary object labels. For this reason, some git commands for naming objects (like git describe) will ignore lightweight tags by default.

### DIFF
--- a/zest/releaser/git.py
+++ b/zest/releaser/git.py
@@ -79,7 +79,7 @@ class Git(BaseVersionControl):
         return ['git', 'log', '%s..HEAD' % version]
 
     def cmd_create_tag(self, version, message, sign=False):
-        cmd = ['git', 'tag', version, '-m', message]
+        cmd = ['git', 'tag', version, '-a', '-m', message]
         if sign:
             cmd.append("--sign")
         if os.path.isdir('.git/svn'):


mauritsvanrees:
Sounds useful. Let me see. I don't explicitly use annotated tags, but when I checkout a tag and do `git describe`, it shows me the tag anyway.

Ah, the same tag documentation says:

> If -m <msg> or -F <file> is given and -a, -s, and -u <keyid> are absent, -a is implied.

We *are* giving a commit message with `-m`, so our tags are actually annotated already.

But what happens when we explicitly pass `-s`, like when you have `[zest.releaser] tag-signing = yes` in `setup.cfg`? Ah, no problem according to the same page:

> Tag objects (created with -a, -s, or -u) are called "annotated" tags; they contain a creation date, the tagger name and e-mail, a tagging message, and an optional GnuPG signature.

This confirms tests that I did just now in a test repo.

Do you agree that the current `zest.releaser` code is already creating annotated tags, and that we can close your PR?
Or am I overlooking something?

kojiromike:
Yes, you are right. Because you are including `-m`, the `-a` is redundant. Thanks for awesome software!

PR #304, opened 2019-01-16
Fix #259: lineendings are properly handled on windows now

### LINKED ISSUE #259: Inserts unnecessary newlines into setup.py and HISTORY.txt
I am a new user to zest releaser. When I run the fullrelease command in a project on Windows, the resulting setup.py and HISTORY.txt has extra newlines inserted between the lines. 

I have not tried running on Linux or MacOS to see if that changes anything.

Before:
```
from setuptools import setup

setup(name='ion_python',
    version='0.1.4.dev0',
    description='desc',
    url='www.example.com',
    author='Kevin Lannen',
    author_email='kevin.lannen@example.com',
    packages=['package_name'],
    zip_save=False,
    test_suite='nose.collector',
    tests_require=['nose'])
```

After:
```
from setuptools import setup



setup(name='ion_python',

    version='0.1.5.dev0',
    description='desc',

    url='www.example.com',

    author='Kevin Lannen',

    author_email='kevin.lannen@example.com',

    packages=['ion_python'],

    zip_save=False,

    test_suite='nose.collector',

    tests_require=['nose'])
```

Running it again seems to insert another set of newlines after each of the lines. Is there a setting that I am missing?

reinout:
We do a `.splitlines()` in `read_text_file()` now. This function uses `rb` mode to get binary data in order to properly detect encodings, so there's no automatic "universal newline" behaviour. With `.splitlines()` we get that behaviour back.

The accompanying `write_text_file()` already uses the regular `w` mode, so "universal newline" behaviour is on already there.

The advantage is that it removes some dirty-looking redefinitions like this:

    lines = read_text_file(...)
    lines = lines.split('\n')


reinout:
Yeah it works now.
The diff I showed earlier was what I got out of "bumpversion" for zest.releaser itself. So bumpversion removed the newline at the end of our own setup.py and changelog.
Now it works, see the version bump I just pushed :-)

mauritsvanrees:
Go ahead and make a release, unless you have other changes planned.

PR #305, opened 2019-01-22
Fix documentation how to configure localization for testing.

htgoebel:
The shell-code was invalid. Also take the chance and remove redundancy in shell-code


reinout:
Thanks for the fix!

(Sometimes I'm thinking about including a docker config for an easier and cleaner test environment, btw).

Issue #310, opened 2019-02-07
bumpversion does not play nice with towncrier


mauritsvanrees:
When a package uses `towncrier` to update the changelog file, the last version header will be something like `5.2.0 (2018-11-09)`. When you call `bumpversion --feature` it will change this to `5.3.0 (2018-11-09)`. That is what happened to me in the [commit in `i18ndude`](https://github.com/collective/i18ndude/commit/2bfc8b659c2d20063b08ed0e82f5629fff081490)

Easiest fix would seem to be: let `bumpversion` only touch the changelog file when there is an `(unreleased)` header in it.
Changing the version in `setup.py` is enough here.

mauritsvanrees:
All fixed by now.

PR #313, opened 2019-02-19
Let bumpversion respect the update_history data.

mauritsvanrees:
Fixes issue #310.
See also https://github.com/collective/zestreleaser.towncrier/pull/11

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,9 @@ Changelog for zest.releaser
 6.16.1 (unreleased)
 -------------------
 
-- Nothing changed yet.
+- Let bumpversion respect the ``update_history`` data.
+  Fixes `issue 310 <https://github.com/zestsoftware/zest.releaser/issues/310>`_.
+  [maurits]
 
 
 6.16.0 (2019-01-17)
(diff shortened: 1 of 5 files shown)


CHANGES.rst
```diff
@@ -4,7 +4,9 @@ Changelog for zest.releaser
 6.16.1 (unreleased)
 -------------------
 
-- Nothing changed yet.
+- Let bumpversion respect the ``update_history`` data.
```
reinout: I only understood this one after reading the pull request. I don't use the "update_history" setting myself, that is probably the cause :-)

Suggestion:

```suggestion
- Let bumpversion respect the ``update_history`` setting.
```

Actually, just so I understand it right: that `update_history` is an *existing* setting that is now used more broadly?

mauritsvanrees: Correct. We introduced this last year so `zestreleaser.towncrier` can tell prerelease and postrelease (and now bumpversion) to not touch the history file.

Maybe as changelog message:

"Let bumpversion not touch the history file when an extension has set ``update_history`` to false."

Would that be clear enough?

reinout: Perhaps be more explicit?

"Better support for zestreleaser.towncrier (and similar extensions): the `update_history` setting is now also respected by the "bumpversion" command.

mauritsvanrees: I have adapted the changelog entry with your suggestion.

Issue #311, opened 2019-02-15
Check if version has 'unreleased' in header


mauritsvanrees:
When your `CHANGES.rst` does not have an `(unreleased)` version, but has for example `1.2.3 (2000-01-01)` as most recent header, then `pre/fullrelease` will find this and happily update it to next version number you are releasing, say `1.2.4 (2019-02-15)`.

This can happen when your package is supposed to use `towncrier` but `zestreleaser.towncrier` is not installed. See this comment in [`plone.restapi`](https://github.com/plone/plone.restapi/pull/618#issuecomment-463971846).

`zest.releaser` can probably check if `(unreleased)` is missing and then either don't touch this file, or give a warning and ask for permission to continue.

mauritsvanrees:
I just realized this is the more general version of issue #310. That one was about bumpversion which does the same.

reinout:
So.... sounds like you smell a solution right around the corner? :-)

PR #314, opened 2019-02-19
Refuse to edit history header when it may be from already released version

mauritsvanrees:
We look for "unreleased" in it.  Give a warning when this happens.
Fixes issue #311.

This is a more general fix than the one in PR #313 which is specific for `bumpversion`. But it seems good to do both.

The current one might possibly have negative side effects when people do not have 'unreleased' in their changelog, but say `(release date to be determined)`. So it could be that this fixes some things and breaks something else. To counter this, we might just need support for an extra item in `setup.cfg`, `unreleased-marker` similar to the `development-marker` that we already have.

But it already fixes something in the tests: in `postrelease.txt` we call `prerelease` and until now this would change `1.0 (1972-12-25)` into `1.3 (<today>)`. With the current PR, it keeps the header, and it would log a warning, though that is not visible in the tests.

### DIFF
--- a/zest/releaser/baserelease.py
+++ b/zest/releaser/baserelease.py
@@ -24,6 +24,7 @@ logger = logging.getLogger(__name__)
 # all commands.
 DATA = {
     'commit_msg': 'Message template used when committing',
+    'has_unreleased_header': 'Latest header is for an unreleased version',
     'headings': 'Extracted headings from the history file',
     'history_encoding': 'The detected encoding of the history file',
     'history_file': 'Filename of history/changelog file (when found)',
@@ -137,6 +138,11 @@ class Basereleaser(object):
             end = len(history_lines)
         history_last_release = '\n'.join(history_lines[start:end])
         self.data['history_last_release'] = history_last_release
+        # Does the latest release header have 'unreleased' in it?
+        # This is useful to know, especially when using tools like towncrier
+        # to handle the changelog.
+        unreleased = 'unreleased' in headings[0]['date']
+        self.data['has_unreleased_header'] = unreleased
 
         # Add line number where an extra changelog entry can be inserted.  Can
         # be useful for entry points.  'start' is the header, +1 is the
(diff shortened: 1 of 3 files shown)


reinout:
You're right. If something else than "unreleased" is somehow used, this PR might break some existing behaviour.

Would it be an idea to do the check the other way around? Try to detect (xxxx-xx-xx)? Just a basic regex `.*\(\d[4]\-\d[2]\-\d[2]\)`-like? 

You wouldn't have false positives this way.

mauritsvanrees:
That could work, although you wouldn't catch the case where people would use a history line like `1.0 (February 20, 2019)`.

reinout:
Using bumpversion is already a bit of a corner case. That is, I've used it only once myself :-)

Supporting "February 20, last year" would be like supporting the corner case of a corner case.

But raising a warning on "not released yet", that's a one-level corner case of regular usage.

mauritsvanrees:
I regularly use `bumpversion` when doing Plone releases.
Note that the other PR (#313) is specifically about `bumpversion`, but this one is also applicable to `prerelease`, as evidenced by the small change in the tests.
I will try the regular expression.

mauritsvanrees:
Fixed.

Test run with a [test package](https://github.com/mauritsvanrees/mauritstestpackage/commit/0b2125431694817177437d81c57d417d5d54557d) for which I already prepared a release yesterday:

```
$ prerelease 
Run pyroma on the package before tagging? (Y/n)? n
Do you want to run check-manifest? (Y/n)? n
Changelog entries for version 1.24.14:

1.24.14 (2019-02-19)
--------------------

- Nothing changed yet.


1.24.13 (2018-03-10)
--------------------
Enter version [1.24.14]: 
WARNING: Changelog contains "- Nothing changed yet.". Are you sure you want to release? (y/N)? y
WARNING: Refused to edit history header, because it looks to be from an already released version with a date. Would have wanted to set this header: 1.24.14 (2019-02-20)
INFO: Set setup.py's version to '1.24.14'
INFO: Changed version from 1.24.14.dev0 to 1.24.14
INFO: History file CHANGES.rst updated.
INFO: The 'git diff':

diff --git a/setup.py b/setup.py
index 4096d10..62a3b10 100644
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,6 @@
 from setuptools import setup, find_packages
 
-version = '1.24.14.dev0'
+version = '1.24.14'
 
 setup(name='mauritstestpackage2',
       version=version,


OK to commit this (Y/n)? 
```

zest/releaser/baserelease.py
```diff
@@ -155,12 +163,32 @@ def _change_header(self, add=False):
...
+        - change a.b.c (unreleased) to a.b.c (today) [prerelease]
+        - change a.b.c (unreleased) to x.y.z (today) [prerelease]
         """
         if self.data['history_file'] is None:
             return
         good_heading = self.data['history_header'] % self.data
+        if not add and self.data.get('has_released_header'):
+            # So we are editing a line, but it already has a release date.
+            logger.warning(
+                'Refused to edit history header, because it looks to be from '
```
reinout: "... edit THE FIRST history header ...": would that be a bit clearer?

Aaargh, we're using "heading" and "header" interchangeably, it seems. I think, headING is more correct.
Perhaps use that in this warning?

mauritsvanrees: Both done.

We seem to use 'header' and 'heading' roughly equal times...

reinout: That's what I feared :-)

reinout:
Releasen maar.

mauritsvanrees:
6.17.0 released.

Issue #315, opened 2019-03-05
Create wheel even when bdist_wheel universal is false


mauritsvanrees:
I have enabled building universal wheels for lots of Plone packages. For a few this causes problems because for those packages the requirements are different between Python 2 and Python 3. Oops, my bad.
So, @gforcada wanted to fix this by switching off universal wheels in `setup.cfg` like [in this commit](https://github.com/plone/plone.app.robotframework/commit/fdbde2ccc936e87ae4db97d219eae08c77442be6):

```
[bdist_wheel]
universal = 0
```

The side effect is that zest.releaser no longer creates wheels at all. See [comment at community.plone.org](https://community.plone.org/t/plone-5-2rc1-soft-released/8163/18?u=mauritsvanrees).
Relevant zest.releaser code is [here](https://github.com/zestsoftware/zest.releaser/blob/6.17.0/zest/releaser/pypi.py#L363-L393).

I would say it is good to change zest.releaser to create a wheel even when universal is false: when there is a bdist_wheel setting, it seems logical that the developer wants a wheel, even if it should not be universal. So we can change our `create_wheel` method to only check that a `bdist_wheel` section is in the `setup.cfg`.

@reinout Do you agree on that?

reinout:
Sounds OK.
Note that I only make universal wheels, so I don't know *that* much of possible additional drawbacks :-)

reinout:
Fixed in #316

PR #316, opened 2019-03-05
When bdist_wheel is in setup.cfg, release a wheel.

mauritsvanrees:
No longer check if this should be a universal wheel or not.
That is handled automatically when calling `python setup.py bdist_wheel`.
You can still set `[zest.releaser] create-wheel = no` to prevent creating a wheel.

This fixes issue #315.


reinout:
I released 6.17.2.

PR #323, opened 2019-04-08
Add support for custom commit messages to the addchangelogentry command.

kleschenko:


mauritsvanrees:
Why is this needed?  The message that you pass as an argument is reused as commit message.  That seems good enough to me.

If you really need this in a script or something, you can do it like this:

```
$ addchangelogentry "Let's foo the bar." && git amend -am "No let's not."
```

kleschenko:
@mauritsvanrees hm, `git commit --amend` will also work for my case, thanks! Forgot about it first and thought that it would be useful to be able to set a different commit message.

mauritsvanrees:
No.

Issue #326, opened 2019-05-23
Feature request: disable the subdirectory warning


wosc:
In one project we have a "monorepo" setup that looks like this:

```
/myrepo
  .git/
  package1/
    setup.py
  package2/
    setup.py
```

Happily, zest.releaser completely does the right thing for `cd package1; fullrelease`, i.e. the egg is correctly created from the `package1` subdirectory and everything. :)

There's just one little niggle, namely the first thing it asks you is "[do you want to go to the root of the repository?](https://github.com/zestsoftware/zest.releaser/blob/master/zest/releaser/choose.py#L29)". Which in this case, no we definitely do not. ;) Do you think that could be supported? I'm not really sure what the appropriate gesture could be, presence of setup.py or setup.cfg or an explicit marker file or something else entirely.

mauritsvanrees:
Seems reasonable, and I guess looking for a `setup.py` would be okay. But it can get tricky. Say you are in `/myrepo/package1/subdir` and you do `fullrelease`.

- The current directory has no `setup.py` file and no `.git` directory, so it is definitely the wrong directory. Fine.
- So we should go one level up, where we find `setup.py` but no `.git`.
- One more and we have `.git` but no `setup.py`.

In this case it should take `/myrepo/.git/` as the repository root, but it should offer to change the directory from `/myrepo/package1/subdir/` to `/myrepo/package1/`.

And then we have a different scenario that I often have, with the same directory structure as you have: one repository for a website, which contains some (buildout) configuration in the top directory and several packages. I don't make individual releases for the packages, but I want a single release for the whole website. So when I am in `/myrepo/package1/subdir` and do `fullrelease`, I definitely *do* want to go to the top level directory.

Now, in my case I always make sure to be in the top level directory when I call `fullrelease`, but the use case is valid. I can imagine that I am in the `/myrepo/package1` directory, do `fullrelease`, and am happy that zest.releaser reminds me that I am not in the repository root: "Yes please, take me to the root, thank you for reminding me." :-)

So the thing is that the `version_control` function does two things:

1. Find the `.git` directory.
2. Possibly change the directory to where a `setup.py` file is or where the `.git` directory is, which may or may not be the same directory.

What we *could* do, is change the default answer to "no" when the current directory has a `setup.py`.

But when we do this properly, we may need to ask not one, but *two* questions:

1. "There is a setup.py file one level up, do you want to go there? (Y/n)"
[...]

PR #328, opened 2019-05-31
Do not go to the root of the repo by default.

mauritsvanrees:
When you were not in the root of a repo, `zest.releaser` already asked if you wanted to go there.
The default answer has now changed from yes to no.
See issue #326.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,10 @@ Changelog for zest.releaser
 6.18.3 (unreleased)
 -------------------
 
-- Nothing changed yet.
+- Do not go to the root of the repo by default.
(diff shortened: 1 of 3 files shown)


CHANGES.rst
```diff
@@ -4,7 +4,10 @@ Changelog for zest.releaser
 6.18.3 (unreleased)
 -------------------
 
-- Nothing changed yet.
+- Do not go to the root of the repo by default.
+  When you were not in the root of a repo, ``zest.releaser`` already asked if you wanted to go there.
+  The default answer has now changed from yes to no.
```
reinout: Perhaps add the reason ("this might help when releasing monorepos")?

mauritsvanrees: OK. I will merge and then add the reason on master.

Issue #333, opened 2019-08-26
addchangelogentry - ValueError exception when commit messages contain % (percent symbol)


dkopelove:
Steps to Reproduce:

1. pip install zest.releaser
2. run command addchangelogentry "%"
3. observe exception

reinout:
I haven't fixed anything yet, but at least I've added a test in #334 that demonstrates that you're right :-)

reinout:
Ah! The actual changelog behaviour worked fine, but the process crashed when adding the commit message. There's a "%" replacement happening in the commit message handling.

I fixed it by doing a quick `%` => `%%` replacement just for the commit message. And I tested it out on my branch, which worked :-)

reinout:
See https://travis-ci.community/t/cannot-install-git-svn-apt-package/863
Travis installs a much newer git (from a PPA) than is normally present in ubuntu xenial.
git-svn doesn't like that, so it is uninstallable.

reinout:
@dkopelove: it should be fixed in 6.19.1!
Thanks for reporting the problem :-)

dkopelove:
Hi @reinout, I just tried 6.19.1 and it works for me as well. Thank you for the quick turn around.

PR #334, opened 2019-09-02
Fix for #333 (percent sign in addchangelogentry command fails)

reinout:


reinout:
See https://travis-ci.community/t/cannot-install-git-svn-apt-package/863
Travis installs a much newer git (from a PPA) than is normally present in ubuntu xenial.
git-svn doesn't like that, so it is uninstallable.

But... hurray, I got it working after much experimenting.
@mauritsvanrees, can you take a look at the fixes in travis.yml?

zest/releaser/tests/utils.txt
```diff
@@ -458,8 +458,9 @@ We want to discover errors and show them in red.
     '\x1b[31m'
     >>> Fore.MAGENTA
     '\x1b[35m'
-    >>> utils.execute_command(['ls', 'some-non-existing-file'])
-    '\x1b[31mls...'
+    >>> output = utils.execute_command(['ls', 'some-non-existing-file'])
+    >>> output.startswith(Fore.RED)
```
reinout: The output somehow started with double quotes, now, instead of single quotes.
That's why I changed the testing style.

PR #340, opened 2020-01-07
Added environment variable ZESTRELEASER

reinout:
This could help #121


reinout:
@mauritsvanrees: I'm setting an env variable in the `__init__` so that a tool like setuptools-scm can pick it up when called from `setup.py -- version` and so. See the bottom of #121

PR #344, opened 2020-02-18
Update README.rst

spereverde:
twine is now a required dependency, 
see https://github.com/zestsoftware/zest.releaser/blob/6.20.0/setup.py


reinout:
Thanks for the fix!
You're right, we need twine now.

PR #347, opened 2020-02-21
Add Dockerfile that allows local tests inside docker.

Behoston:
Now I'm able to run tests locally and don't care about system dependencies inside docker container.
I hope that will help new contributors.

obraz


reinout:
Here's my PR with textual improvements: https://github.com/zestsoftware/zest.releaser/pull/348
Can you take a look?

Issue #351, opened 2020-04-13
Unable to choose repository for upload automatically


LvffY:
Hello, 

First, thanks your all your work :)

I'm kind of a newbie in Python package releases but from what I saw from my other experiences, your project is very simple to use and effective ! :)

Here is my problem : In my case, I could have many private PyPi. Actually, our purpose is to have 1 PyPi repository per projet grouped in 1 big group nexus repository for customers entry point. 

For this, I have setup my `~/.pypirc' file like this : 

```ini
[distutils]
index-servers =
	pypi-project1
	pypi-project2

[pypi-project1]
repository=http://myServer1/repository/PROJECT1_RELEASE/
username=user1
password=pwd1

[pypi-project1]
repository=http://myServer2/repository/PROJECT2_RELEASE/
username=user2
password=pwd2
```

For now, with manual `fullrelease` command, I'm always asked for each repository if I want to upload my codes. Is there any configurations that I missed that could, in case of `--no-input` option, automatically upload to only one of my repository ? 

Sorry, if I missed an issue discussing about this and thanks in advance for your responses !

LvffY:
Apologize for not reading the docs, I found my solution ... 

If anyone fall into this problem, I solved my problem by : 

- Having an `~/.pypirc` file like this : 

```ini
[distutils]
index-servers =
	pypi-project1
	pypi-project2

[pypi-project1]
repository=http://myServer1/repository/PROJECT1_RELEASE/
username=user1
password=pwd1

[pypi-project1]
repository=http://myServer2/repository/PROJECT2_RELEASE/
username=user2
password=pwd2
```

- A `setup.cfg` file at the root of my project looking like : 

```ini
[distutils]
index-servers =
	pypi-project1
```

I found the solution by reading (more properly) the [docs](https://zestreleaser.readthedocs.io/en/latest/uploading.html#pypi-configuration-file-pypirc) especially : 

« 
Note that since version 3.15, zest.releaser also looks for this information in the setup.cfg if your package has that file. One way to use this, is to restrict the servers that zest.releaser will ask you to upload to. If you have defined 40 index-servers in your pypirc but you have the following in your setup.cfg, you will not be asked to upload to any server:

```ini
distutils]
index-servers =
```
»

By analogy, I solved my own problem. *May be another example could be a good idea to image this problem ?*

Anyway, sorry for the issue that wasn't one ... And once again, thanks for your work ! :)

reinout:
[...]

PR #352, opened 2020-04-14
Add example for restricting the used index servers.

mauritsvanrees:
https://github.com/zestsoftware/zest.releaser/issues/351


reinout:
I like the change, but travis complains loudly. 2.7 smells like we need an older twine on 2.7.
Do you recognize the 3.8 buildout failure? I've been kinda ignoring the recent buildout error reports as I rarely use it these days...

reinout:
Note: I'm playing with the thought of setting up "github actions" instead of travis. I'm using it for a few work projects now: they spin up real fast.
A docker could also be a nice addition. I'm having problems getting all the tests to pass on my own laptop right now :-)

mauritsvanrees:
It works now locally (on Python 2.7 which I focused on now for the versions), except that the git-svn tests fail, because git-svn itself no longer works on my Mac.
Well, I have a few test failures in `utils.txt` where there is a confusion between an expected SystemExit and a RuntimeError. But `bin/test -t utils.txt` works fine.

Anyway, Travis finally passes.

Switching the tests to use `pip install` and run `pytest` or something like that would be fine for me as well. We now have some automatic grokking.

reinout:
Hurray that you got it working!

PR #354, opened 2020-06-12
353 - Add support for Twine environment variables

mctwynne:
Twine supports a number of environment variables which can be
used to override settings in setup.cfg and pypirc. Using the
TWINE_REPOSITORY and TWINE_REPOSITORY_URL environment variables
can now be used to override the desired index server and/or URL
that packages wil be uploaded to.


mctwynne:
@mauritsvanrees I'm getting `SyntaxError`s in the 2.7 and pypy CI jobs, but they don't *seem* to be related to my change. Is this something you've seen before, or do you happen to know what I've broken? Thanks!

mauritsvanrees:
On Python 2.7 we are getting a too new `check-manifest` which only works on Python 2.7.
Locally I actually get much more errors when I try to run the buildout on 2.7. I will have a look.

mauritsvanrees:
My other PR is merged, so if you rebase your branch on master (or merge master), Travis should be fixed.

mctwynne:
@mauritsvanrees Twine seems to ignore any command line arguments when the environment variables are present. It does not actually fail if only `TWINE_REPOSITORY_URL` is set, since it's passed as `-r <URL>` but that is ignored.

In the case where only `TWINE_REPOSITORY` is set to a URL or a server name which does not exist in `.pypirc` twine will fail:

```
twine.exceptions.InvalidConfiguration: Missing 'SERVER NAME OR URL' section from the configuration file
or not a complete URL in --repository-url.
```
If both environment variables are set, but the `TWINE_REPOSITORY` does not exist in `.pypirc`, it's also ignored and the `TWINE_REPOSITORY_URL` is used.

I didn't change the way twine uploads were being done, as this change seemed to be the least invasive. That being said, it is somewhat unintuitive.

I'd love to have some tests for this, but I'm not exactly sure how (or if it even makes sense) to include mocks in the doctests. Mocks are also incompatible with 2.7 from what I can tell (?), which is unfortunate. Actually setting environment variables in the tests is also not a great idea. Any thoughts on this?

buildout.cfg
```diff
@@ -30,7 +30,7 @@ bleach = 2.1.2
 certifi = 2018.1.18
 cffi = 1.13.2
 collective.recipe.sphinxbuilder = 1.0
-colorama = 0.3.9
+colorama = 0.4.3
```
mctwynne, line 33: @mauritsvanrees I needed to update this to get the Docker image to build for tests.

mctwynne:
@mauritsvanrees Thanks for accepting this change! Glad to hear it's working for you :)

PR #356, opened 2020-06-23
Pin more versions on Python 2.7 (and PyPy).

mauritsvanrees:
I hope this fixes Travis.
We get too new versions of packages, that are not compatible.

Works locally for Python 2.7 and PyPy.

(`git svn` tests are failing because I don't have this properly installed anymore it seems. And only on PyPy a few mercurial tests fail because of a missing user name, which is not the case on 2.7. But this is both likely only a local problem.)


reinout:
You can merge it once it is finished. Changes look good.

mauritsvanrees:
Green!

PR #361, opened 2020-09-02
Fix TypeError on Python 2.7 and git-lfs issue

mcdeck:
I bumped into a couple of issues using the latest version of zest.releaser under python 2.7, while trying to release code that uses git and git-lfs.

This PR address both these issues.

### DIFF
--- a/zest/releaser/baserelease.py
+++ b/zest/releaser/baserelease.py
@@ -57,7 +57,7 @@ NOTHING_CHANGED_YET = '- Nothing changed yet.'
 class Basereleaser(object):
 
     def __init__(self, vcs=None):
-        os.environ["ZESTRELEASER"] = "We are called from within zest.releaser"
+        os.environ[str("ZESTRELEASER")] = str("We are called from within zest.releaser")
         # ^^^ Env variable so called tools can detect us. Don't depend on the
         # actual text, just on the variable's name.
         if vcs is None:
--- a/zest/releaser/utils.py
+++ b/zest/releaser/utils.py
@@ -764,8 +764,6 @@ def _subprocess_open(p, command, input_value, show_stderr):
         if stderr_output:
             logger.debug("Stderr of running command '%s':\n%s",
                          format_command(command), stderr_output)
-    o.close()
-    e.close()
     return result
 
 
--- a/zest/releaser/utils.py
+++ b/zest/releaser/utils.py
@@ -741,11 +741,10 @@ __command_is_string__ = False
 
 
 def _subprocess_open(p, command, input_value, show_stderr):
-    i, o, e = (p.stdin, p.stdout, p.stderr)
     if input_value:
-        i.write(input_value.encode(INPUT_ENCODING))
-    i.close()
-    (stdout_output, stderr_output) = p.communicate()
+        (stdout_output, stderr_output) = p.communicate(input_value.encode(INPUT_ENCODING))
+    else:
+        (stdout_output, stderr_output) = p.communicate()
     # We assume that the output from commands we're running is text.
     if not isinstance(stdout_output, six.text_type):
         stdout_output = stdout_output.decode(OUTPUT_ENCODING)
--- a/zest/releaser/utils.py
+++ b/zest/releaser/utils.py
@@ -745,8 +745,7 @@ def _subprocess_open(p, command, input_value, show_stderr):
     if input_value:
         i.write(input_value.encode(INPUT_ENCODING))
     i.close()
-    stdout_output = o.read()
-    stderr_output = e.read()
+    (stdout_output, stderr_output) = p.communicate()
     # We assume that the output from commands we're running is text.
     if not isinstance(stdout_output, six.text_type):
         stdout_output = stdout_output.decode(OUTPUT_ENCODING)


zest/releaser/baserelease.py
```diff
@@ -57,7 +57,7 @@
 class Basereleaser(object):
 
     def __init__(self, vcs=None):
-        os.environ["ZESTRELEASER"] = "We are called from within zest.releaser"
+        os.environ[str("ZESTRELEASER")] = str("We are called from within zest.releaser")
```
mauritsvanrees, line 60: What does this fix? I don't think this changes anything in any Python version.
I tried the original in Python 2.7, 3.7, 3.9, pypy2, pypy3: all worked fine.
So simply this:

```
$ python
>>> import os
>>> os.environ["ZESTRELEASER"] = "We are called from within zest.releaser"
```

zest/releaser/utils.py
```diff
@@ -745,8 +745,7 @@ def _subprocess_open(p, command, input_value, show_stderr):
     if input_value:
         i.write(input_value.encode(INPUT_ENCODING))
     i.close()
-    stdout_output = o.read()
-    stderr_output = e.read()
+    (stdout_output, stderr_output) = p.communicate()
```
mauritsvanrees, line 748: This fails on all Python versions on Travis:

    ValueError: I/O operation on closed file

We could certainly be something weird and non-standard in our code. This part has been tweaked various times, trying to keep it working in various Python versions. It may not be the most modern way to do this.

Does this actually fix anything for you in practice? Or is this just a try?

Just a guess: maybe if there is an `input_value`, you should pass it to `p.communicate`. So the beginning of the function would look like this:

```
def _subprocess_open(p, command, input_value, show_stderr):
    if input_value:
        (stdout_output, stderr_output) = p.communicate(input_value.encode(INPUT_ENCODING))
    else:
        (stdout_output, stderr_output) = p.communicate()
```

Note: I have not tried this.

zest/releaser/baserelease.py
mcdeck: I actually got this stack trace when running `release -v` on my repo:
```Traceback (most recent call last):
  File "C:\Python27-64\lib\runpy.py", line 174, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "C:\Python27-64\lib\runpy.py", line 72, in _run_code
    exec code in run_globals
  File "C:\Development\sources\stage\_venv\Scripts\release.exe\__main__.py", line 7, in <module>
  File "c:\development\sources\stage\_venv\lib\site-packages\zest\releaser\release.py", line 346, in main
    releaser = Releaser()
  File "c:\development\sources\stage\_venv\lib\site-packages\zest\releaser\release.py", line 63, in __init__
    baserelease.Basereleaser.__init__(self, vcs=vcs)
  File "c:\development\sources\stage\_venv\lib\site-packages\zest\releaser\baserelease.py", line 70, in __init__
    'name': self.vcs.name,
  File "c:\development\sources\stage\_venv\lib\site-packages\zest\releaser\git.py", line 31, in name
    package_name = self.get_setup_py_name()
  File "c:\development\sources\stage\_venv\lib\site-packages\zest\releaser\vcs.py", line 93, in get_setup_py_name
    utils.execute_command(utils.setup_py('egg_info'))
  File "c:\development\sources\stage\_venv\lib\site-packages\zest\releaser\utils.py", line 866, in execute_command
    result = _execute_command(command)
  File "c:\development\sources\stage\_venv\lib\site-packages\zest\releaser\utils.py", line 810, in _execute_command
    process = subprocess.Popen(command, **process_kwargs)
  File "C:\Python27-64\lib\subprocess.py", line 394, in __init__
    errread, errwrite)
  File "C:\Python27-64\lib\subprocess.py", line 644, in _execute_child
    startupinfo)
TypeError: environment can only contain strings```

And I traced that back to having this in the the Zest release as a unicode object in the environment passed to subprocess:
```
{ ...
u'ZESTRELEASER: u'We are called from within zest.releaser'
... }
```

And I figured that was due to a combination of using `unicode_literals` on Python 2.7 (Python 2.7.15, to be precise). It may also be some or another oddity on my installation, because I also did not see that specific problem on another installation.

Anyway, forcing this into 'str' type fixed the issue for me.

File: zest/releaser/utils.py (hunk 2 above)
Comment (mcdeck) [reply zest/releaser/utils.py:748]: I'll give this a try with what you suggested. 

It did in fact cause a problem for me when running zest.releaser in a git repo that uses git-lfs. The process actually deadlocks in that case while running 
```
DEBUG: Running command: 'git clone --depth 1 C:\Development\sources\stage gitclone'
```

With this change the deadlock is gone. I thought it may be related to what the python doc has to say about `Popen`, `PIPE`, et al (https://docs.python.org/2/library/subprocess.html):

> Note:  This will deadlock when using stdout=PIPE and/or stderr=PIPE and the child process generates enough output to a pipe such that it blocks waiting for the OS pipe buffer to accept more data. Use communicate() to avoid that.

mcdeck: For the record, the fix worked for my specific situation, but I of course forgot to run any test locally before opening the PR, sorry for that.

zest/releaser/baserelease.py
mauritsvanrees: Ah, I did not realise we are using `unicode_literals`. I don't think I have ever used them, but I see we use them in `zest.releaser`.
Strange: even with unicode literals it works fine for me on Python 2.7. I am on a Mac. Maybe only Windows has a problem with this.

But I do expect a standard string here, so forcing it to a string like you do is certainly good.

zest/releaser/utils.py
mauritsvanrees: So: it could be that until now it has worked fine (at least for me) only because there was not much output. Interesting.
@reinout You are using `git-lfs` something I think. Have you seen this problem?

The tests now fail with `NameError: global name 'o' is not defined`. So a bit more cleanup is needed in that function.

zest/releaser/baserelease.py
reinout: I see a similar report here: https://github.com/peritus/bumpversion/issues/135 and here: https://github.com/theskumar/python-dotenv/pull/101

Very weird that it didn't occur earlier?

zest/releaser/utils.py
reinout: I used git-annex for a while, not git-lfs.

mcdeck: > The tests now fail with `NameError: global name 'o' is not defined`. So a bit more cleanup is needed in that function.

Sorry for that - I was in a hurry this morning and did (again) not test things, reasoning that "what could go wrong", and here we are :)
I'll get CI green in the course of this evening.

mauritsvanrees (approved):
Tests are green. I did a test release with your code, and it works. Looks cleaner now.
Changelog entry is missing, but I will add something after merge.
Thanks!

mauritsvanrees:
I have released 6.22.0 with your fixes. Thanks again!

PR #372, opened 2021-07-05
Small setup fix (was: allow tools prefix)

### LINKED ISSUE #371: support [tool:zest.releaser] in setup.cfg
tools like pytest  use a header [tool:pytest]  in setup.cfg

`
[tool:isort]
....

[tool:pep8]
`

but zest.releaser breaks on this:  
the setting  python-file-with-version = src/mypackage/__init__.py  is found with header [zest.releaser], 
and not found when using  [tool:zest.releaser],

reinout:
Originally started for #371, but that looks like too much work.

What remains is that I've fixed up the build setup slightly.


reinout:
See my comment on #371: I think it is more work than I originally thought. Not really worth it.

@mauritsvanrees : the local build fixes, do these look OK to you?

mauritsvanrees:
@reinout Yes, those look fine. `bootstrap.py` is dead.

Also, we should move to GitHub Actions instead of Travis at some point, but for the moment it should still work.

mauritsvanrees:
`.travis.yml` roept nog `bootstrap.py` aan.

reinout:
I've removed bootstrap from .travis.yml and 3.8 runs fine.
But for instance 3.7 has this error when pip-installing zc.buildout and setuptools:

    zc-buildout 2.13.4 has requirement setuptools<52,>=8.0, but you'll have setuptools 57.1.0 which is incompatible.

I haven't used buildout for a while. Does this make sense? Perhaps I shouldn't have upgraded setuptools with pip?

reinout:
More success now (3.6, 3.7 and 3.8), but the rest has problems with pyroma.
What to do about that?

mauritsvanrees:
Pin it to an older version on older Python versions. I have pushed a commit for that.

reinout:
Yeah, that worked :-)

PR #377, opened 2021-11-04
Black

mauritsvanrees:
@reinout Do you have a preference about the quotes? I prefer double quotes, which is what `black` without any options does. Done in the second commit.


reinout:
Double quotes please, yes :-) Otherwise `error_message = "Don't you dare touch this"` is such a bother.

mauritsvanrees:
I prefer double quotes too. Too bad that the Python interpreter prefers single quotes:

```
>>> "double quotes"
'double quotes'
```

PR #378, opened 2021-11-05
Removed Python 2 fallbacks.

mauritsvanrees:
The main ones:

- Always use tokenize.open instead of chardet to detect the encoding.
  We meant chardet for Python 2, although theoretically it might
  be needed in corner cases on Python 3.  Let's try without it.
- Use shlex.quote instead of our simpler quoting in format_command.


reinout (approved):
I'm wondering a little bit about removing the encoding hints stuff, but yeah, big chance that it isn't needed anymore.
So: yeah, looks good!

PR #379, opened 2021-11-16
Use subprocess.run for executing commands

mauritsvanrees:
If I recall correctly, our `_execute_command` function started out as a copy of some code from `zc.buildout`, with as comment that it was a "commands.getoutput() replacement that also works on windows".
Meanwhile, the entire [`commands` module was removed in Py 3](https://docs.python.org/2.7/library/commands.html?highlight=commands#module-commands)  and you should use `subprocess` now.

We already did this, but we were using `subprocess.Popen`. The [`subprocess` documentation says](https://docs.python.org/3.10/library/subprocess.html):

> The recommended approach to invoking subprocesses is to use the run() function for all use cases it can handle. For more advanced use cases, the underlying Popen interface can be used directly.
> The run() function was added in Python 3.5

It looks like our use cases can be handled just fine by `run`, so I did that.
And I got rid of some more encoding-related and Windows-specific code, with the idea that standard Python probably knows better how to handle this than we do.

The amount of code changed in this PR is not so much, but it covers some complicated code that I don't usually touch. It may help to look at the smaller individual commits that simplify the code step by step.

### DIFF
--- a/zest/releaser/utils.py
+++ b/zest/releaser/utils.py
@@ -18,9 +18,6 @@ import tokenize
 logger = logging.getLogger(__name__)
 
 WRONG_IN_VERSION = ["svn", "dev", "("]
-# For zc.buildout's system() method:
-MUST_CLOSE_FDS = not sys.platform.startswith("win")
-
 AUTO_RESPONSE = False
 VERBOSE = False
 INPUT_ENCODING = "UTF-8"
@@ -677,7 +674,7 @@ def _subprocess_open(p, command, show_stderr):
 
 
 def _execute_command(command):
-    """commands.getoutput() replacement that also works on windows"""
+    """Execute a command, returning stdout, plus maybe parts of stderr."""
     # Enforce the command to be a list or arguments.
     assert isinstance(command, (list, tuple))
     logger.debug("Running command: '%s'", format_command(command))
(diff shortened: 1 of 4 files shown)


zest/releaser/utils.py
```diff
@@ -695,8 +695,8 @@ def _execute_command(command):
         "stderr": subprocess.PIPE,
         "env": env,
     }
-    with subprocess.Popen(command, **process_kwargs) as process:
-        return _subprocess_open(process, command, show_stderr)
+    process = subprocess.run(command, **process_kwargs)
+    return _subprocess_open(process, show_stderr)
```
reinout: I'm wondering a bit about still using the `_subprocess_open()` name. Perhaps it is just a naming thing.
And probably you've fixed it in another commit (I'm going through everything commit-by-commit as you suggested :-) )

zest/releaser/utils.py
```diff
@@ -650,27 +647,19 @@ def format_command(command):
 
 
 def _subprocess_open(process, show_stderr):
```
reinout: Yeah, the name could become "format output" or "filter output" or so.

zest/releaser/utils.py
```diff
@@ -641,22 +641,6 @@ def format_command(command):
     return " ".join(args)
 
 
-def _subprocess_open(process, show_stderr):
```
reinout: Removing the function name solves everything.

reinout (approved):
I had some comment halfway, but by the end it was solved.

PR #389, opened 2022-09-07
🐛 Add option to add message before the commit messages for zest.releaser 7.x

LvffY:
Closes #388

PR #390, opened 2022-09-07
🐛 Add option to add message before the commit messages for zest.releaser 6.x

### LINKED ISSUE #388: Possibility to add prefix to commit messages
Hello

I would like to be able to more configure our release process by adding a prefix message to our commit messages. *I know about `extra-message` configuration, but this add message at the end of the commit, here I want the message to be <b>before</b> the commit message*

For example, I think about a `prefix-message` configuration that could be added in the configuration files. 

That would lead to : 

- **setup.cfg**

```ini
[zest.releaser]
prefix-message = [This is a message before]
extra-message = [This is a message after] ## Only for the example, not needed
```

- The commit message would then be : 

```
[This is a message before] Preparing release 1.0.3.0

[This is a message after]
```

Do you think this could be done ? I'd be happy to submit a PR if you think this is useful.

LvffY:
Closes #388

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,8 +4,13 @@ Changelog for zest.releaser
 6.22.3 (unreleased)
 -------------------
 
-- Nothing changed yet.
-
+- Bug 388: Optionally add prefix text to commit messages.  This can be used to
+  avoid running Travis Continuous Integration builds.  See
+  http://docs.travis-ci.com/user/how-to-skip-a-build/.  To activate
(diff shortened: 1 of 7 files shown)


LvffY:
@reinout this follows up #389 for zest.releaser 6.x versions

reinout:
I'm pretty sure we won't release a 6.x version anymore. At least, @mauritsvanrees, I don't think we will?
Anyway, high time to remove the alpha marker from 7.x :-)

PR #398, opened 2022-12-09
Support Markdown for History/Changelog

ericof:
- [X] Add `history_format` option
- [X] Automatically detect history_format based on filename extension
- [X] Update GHA workflow to the latest versions


mauritsvanrees (approved):
Tested and it works. Thanks!

As I wrote on Discord, instead of hardcoding `##`, nicer would be if we detect what the header level should be, by checking the other levels, like we currently do for seeing what the RST underlines should be. But that is nice for later.

reinout:
Great feature, thanks!

Issue #373, opened 2021-08-21
support pep621


graingert:
pep621 is now publicly supported and documented in flit, zest.releaser should support it also by setting the pyproject.toml :

```toml
[project]
version = "0.0.0a0"
```
field

(While preserving quote type too)

graingert:
Also if zest sees this field it should take priority over any version in any other location, eg setup.py/cfg

regebro:
I looked into this a bit more, and I think supporting this doesn't really make any sense unless zest.releaser also supports building the package with flit, so in essence I think this really is fully a part of #295.

graingert:
Setuptools 61 now supports PEP 621 (and setup.cfg is pending deprecation)

mauritsvanrees:
Version 8.0.0a1 supports writing the version to `pyproject.toml`.
For releasing you still need a `setup.py` for now. currently `zest.releaser` simply checks if there is a `setup.py`. If it is not there, `zest.releaser` does not create distributions, and does not push to PyPI. That will hopefully be for later in the alpha stage.

Issue #295, opened 2018-11-07
Implement PEP-518


xsc27:
Having the ability to use pyproject.toml instead of setup.cfg would be a "nice to have" feature.

reinout:
Yes, agreed.
The code is already pretty flexible where it grabs its settings, so an extra file to look in can be added reasonably simply.

Only thing to watch out for is whether we need an extra dependency to support the .toml syntax. I assume that's OK.

graingert:
usually projects use a `[toml]` extra

WouterVH:
any progress? 
Trying to move my tools-config to pyproject.toml.

graingert:
PEP 621 support would be even better!

mauritsvanrees:
I don't have an immediate need for this, so I am not likely to work on this in the near future.  But if someone wants to make a PR, that would be fine.

Question is which toml parser to use. @graingert mentions [`tomli`](https://pypi.org/project/tomli/), but [`toml`](https://pypi.org/project/toml/) has more stars, and is for example used by `towncrier`, which I use in the `zestreleaser.towncrier` extension, so I would be inclined to use `toml`.

I wonder: what would we do with it? Only use this to read the `zest.releaser` configuration options?

Or do we also for example read and write the version number in `pyproject.toml`? I have not used this file except for towncrier setup, so I don't know if this is supported by one of the dozen PEPs about version numbers...

When we read config options, I would say the following is a good order, where the first file where an option is found, wins:

- command line option 
- `pyproject.toml`
- `setup.cfg`
- `~/.pypirc`

WouterVH:
>> Only use this to read the zest.releaser configuration options?
 Yes.  Newer tools like black only support pyproject.toml and no setup.cfg.
Several tools that read their settings from setup.cfg, have not supported for pyproject.toml, like towncrier, check-manifest, isort,...

graingert:
toml doesn't support toml v1 syntax.  pip is also moving to tomli. https://github.com/pypa/pip/pull/10035

regebro:
In the long run I suppose it needs to support updating the version no matter if it is in setup.cfg or pyproject.toml.

But looking into pyproject.toml at the moment, it seems like setuptools doesn't fully support it. I think the only build project that supports not having a setup.cfg is Poetry?

graingert:
And flit

graingert:
Projects have now for the most part switched to tomli

graingert:
[...]

PR #401, opened 2023-02-08
Write version in pyproject toml if available

mauritsvanrees:
- Added `pyproject.toml` and moved most metadata there. "Eat your own dog food." I am not sure if we can get rid of the remaining parts of `setup.py`.
- Add reader. I added `tomli` to the dependencies on Pythons earlier than 3.11. Chances are that you already have this via a different dependency. On 3.11 I use the built-in `tomllib`. 
- Add writer. This is not that different from how we write to `setup.py` or other files.
- Checked that `bin/prerelease` updates the version in our own `pyproject.toml`. I did not commit this.

This fixes issue #373.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,8 @@
+[build-system]
+# See https://snarky.ca/what-the-heck-is-pyproject-toml/
+requires = ["setuptools", "wheel"]
+build-backend = "setuptools.build_meta"
+
 [project]
 name = "zest.releaser"
 version = "7.3.1.dev0"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,6 +27,7 @@ classifiers = [
     "Programming Language :: Python :: 3.8",
     "Programming Language :: Python :: 3.9",
     "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: Implementation :: CPython",
     "Programming Language :: Python :: Implementation :: PyPy",
     "Topic :: Software Development :: Libraries :: Python Modules",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,6 +12,7 @@ dependencies = [
     "colorama",
     "requests",
     "twine >= 1.6.0",
+    "tomli; python_version<'3.11'",
 ]
 requires-python = ">=3.6"
 classifiers = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,7 +19,7 @@ dependencies = [
     "twine >= 1.6.0",
     "tomli; python_version<'3.11'",
 ]
-requires-python = ">=3.6"
+requires-python = ">=3.7"
 classifiers = [
     "Development Status :: 6 - Mature",
     "Intended Audience :: Developers",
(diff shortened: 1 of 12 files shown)


pyproject.toml
```diff
@@ -1,2 +1,91 @@
...
+    "Intended Audience :: Developers",
+    "License :: OSI Approved :: GNU General Public License (GPL)",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3 :: Only",
+    "Programming Language :: Python :: 3.6",
+    "Programming Language :: Python :: 3.7",
+    "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
```
reinout, line 29: Add 3.11 too?

setup.py
```diff
@@ -26,88 +23,9 @@ def read(filename):
...
-        "Programming Language :: Python :: Implementation :: CPython",
-        "Programming Language :: Python :: Implementation :: PyPy",
-        "Topic :: Software Development :: Libraries :: Python Modules",
-    ],
-    keywords=["releasing", "packaging", "pypi"],
-    author="Reinout van Rees",
-    author_email="reinout@vanrees.org",
-    url="https://zestreleaser.readthedocs.io",
-    license="GPL",
     packages=find_packages(exclude=["ez_setup"]),
```
reinout, line 27: We can probably zap `ez_setup` altogether? Wasn't that sort of a setuptools bootstrap file for when setuptools was new? I don't totally remember :-)

zest/releaser/vcs.py
```diff
@@ -233,6 +249,50 @@ def _update_python_file_version(self, version):
         utils.write_text_file(filename, contents, encoding)
         logger.info("Set __version__ in %s to '%s'", filename, version)
 
+    def _update_pyproject_toml_version(self, version):
```
reinout, line 252: Test?

reinout (commented):
Good idea!
Most important point: I think a test has to be added? There will be corner cases later on, I fear.

pyproject.toml
mauritsvanrees: Done.

setup.py
mauritsvanrees: Right, it only makes sense to exclude `ez_setup.py` if we actually have this file, which we don't. I removed the code.

zest/releaser/vcs.py
mauritsvanrees: Done. See commit 1bd131523698c383d6974ded8c55a95355b498c6.

mauritsvanrees:
Okay, so I added a test, and it works fine. The message from the commit may be interesting: 1bd131523698c383d6974ded8c55a95355b498c6.

Problem: the test does not work on Python 3.6, so I dropped support for that. We will want to make a major release for this.
And the test shows that if you *only* have a `pyproject.toml` and no `setup.py` at all, then `fullrelease` does not create and upload distributions to PyPI. So that is something that we need to fix. That is basically issue #295.

I think the current PR is fine to merge: it enables updating the version when it is in `pyproject.toml`. But releasing still needs a `setup.py` at this point, though it can be mostly empty.
We could make an 8.0.0a1 release with this.

reinout:
I agree. I'll merge it.

Issue #402, opened 2023-02-08
tests fails with non-GNU ls


mtelka:
Example of failure:
```
File "$(BUILD_DIR)/zest/releaser/tests/utils.txt", line 554, in utils.txt
Failed example:
    print(result)
Expected:
    RED ls...
Got:
    RED some-non-existing-file: No such file or directory
```
The root cause seems to be this:
```
$ /bin/ls /x
/x: No such file or directory
$ /usr/gnu/bin/ls /x
ls: cannot access '/x': No such file or directory
$
```
OS: OpenIndiana

reinout:
Hey, that's a new one :-) We used to have problems with variations in output of git, svn, hg. I didn't expect one in `ls`.

The easiest way is probably to run the tests with docker (see https://zestreleaser.readthedocs.io/en/latest/developing.html ), that's useful to work around non-critical deviations like this.

Alternatively, the tests can probably be adjusted to expect `RED ... No such file or directory`, but using the docker is less work. I assume zest.releaser as such works OK on openindana?

mauritsvanrees:
We also have test output normalisation. We could strip 'ls:' there.

reinout:
That's also a possibility. But it also depends on how many differences there are. @mtelka, is it just "ls" or are there more differences?

mtelka:
I'm packaging zest.releaser for OpenIndiana so running tests in docker is no way.  I need to run tests in real and current environment to make sure the created package is okay.

There are more test failures than this one, but I need to analyze them deeper to find where problems really are.  The ls output difference is just so obvious so I reported it.

PR #403, opened 2023-02-08
Remove 'ls' from expected test output.

mauritsvanrees:
Apparently this is not printed when you use the OpenIndiana OS. Fixes https://github.com/zestsoftware/zest.releaser/issues/402

There is only one test file where we need to fix this, and I think it is fine to ignore a few more characters here.

@mtelka Are you able to try this branch?


reinout (approved):

mtelka:
I tested 9d3eae646eecb72474dcf87a2a5016e3a2728b20 and it hepled.  Thank you.

mauritsvanrees:
I have merged this and released 8.0.0a1. This should also have the `tox.ini` included.

Issue #408, opened 2023-04-06
get_last_tag can fail with recent setuptools


mauritsvanrees:
This is with zest.releaser 8.0.0a1, though I guess earlier versions are affected as well.

`setuptools` 66 is more strict with what versions it can parse.
With a checkout of [`plone.app.event`](https://github.com/plone/plone.app.event) I get this:

```
$ lasttaglog
Traceback (most recent call last):
  File "/Users/maurits/community/plone-coredev/6.0/src/plone.app.event/../../bin/lasttaglog", line 67, in <module>
    sys.exit(zest.releaser.lasttaglog.main())
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/maurits/shared-eggs/cp311/zest.releaser-8.0.0a1-py3.11.egg/zest/releaser/lasttaglog.py", line 22, in main
    found = utils.get_last_tag(vcs)
            ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/maurits/shared-eggs/cp311/zest.releaser-8.0.0a1-py3.11.egg/zest/releaser/utils.py", line 856, in get_last_tag
    parsed_tag = parse_version(tag)
                 ^^^^^^^^^^^^^^^^^^
  File "/Users/maurits/community/plone-coredev/6.0/lib/python3.11/site-packages/pkg_resources/_vendor/packaging/version.py", line 197, in __init__
    raise InvalidVersion(f"Invalid version: '{version}'")
pkg_resources.extern.packaging.version.InvalidVersion: Invalid version: 'PLIP-2780-PRE-ES6'
```

That is [this tag](https://github.com/plone/plone.app.event/releases/tag/PLIP-2780-PRE-ES6).

Same for `lasttagdiff` and `bumpversion`.

If `parse_version` fails, we should ignore this tag.  Maybe warn.

mauritsvanrees:
Fix released in 8.0.0a2.

PR #409, opened 2023-04-06
Do not fail when tag versions cannot be parsed.

mauritsvanrees:
This can happen in `lasttaglog`, `lasttagdiff`, and `bumpversion`, with `setuptools` 66 or higher. Fixes issue #408.

I see a few test failures locally that should be unrelated.  The tests expect a master branch, but there is a main branch.  This probably differs per git version or local configuration.  Let's see what gh-actions says.

### DIFF
--- a/zest/releaser/utils.py
+++ b/zest/releaser/utils.py
@@ -849,19 +849,32 @@ def get_last_tag(vcs, allow_missing=False):
     # by setuptools' version parsing. A direct match is obviously
     # right. The 'less' approach handles development eggs that have
     # already been switched back to development.
-    available_tags.reverse()
-    found = available_tags[0]
+    # Note: if parsing the current version fails, there is nothing we can do:
+    # there is no sane way of knowing which version is smaller than an unparsable
+    # version, so we just break hard.
     parsed_version = parse_version(version)
+    found = parsed_found = None
     for tag in available_tags:
-        parsed_tag = parse_version(tag)
-        parsed_found = parse_version(found)
+        try:
+            parsed_tag = parse_version(tag)
+        except Exception:
+            # I don't want to import this specific exception,
+            # because it sounds unstable:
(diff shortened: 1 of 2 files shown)


mauritsvanrees:
Okay, tests passed on gh-actions, so may be something local.
Still, I have added a commit to call `git init -b main` when setting up the example package in the tests.  Now tests work fine locally again.

Well, there are a few test failures because I have this in my `.pypirc`:

```
[zest.releaser]
extra-message =
    [ci skip]
```

But never mind that for now.

zest/releaser/utils.py
```diff
@@ -849,19 +849,32 @@ def get_last_tag(vcs, allow_missing=False):
     # by setuptools' version parsing. A direct match is obviously
     # right. The 'less' approach handles development eggs that have
     # already been switched back to development.
-    available_tags.reverse()
```
mauritsvanrees, line 852: Note: I have removed this line that reverses the available tags.  We want to go through the entire list anyway, so I don't think it matters.

zest/releaser/utils.py
```diff
@@ -849,19 +849,32 @@ def get_last_tag(vcs, allow_missing=False):
...
     for tag in available_tags:
-        parsed_tag = parse_version(tag)
-        parsed_found = parse_version(found)
+        try:
+            parsed_tag = parse_version(tag)
+        except Exception:
+            # I don't want to import this specific exception,
+            # because it sounds unstable:
+            # pkg_resources.extern.packaging.version.InvalidVersion
+            continue
```
reinout, line 864: Would a `logger.debug(e)` be handy here for debugging if something goes wrong with versions?

reinout (approved):

zest/releaser/utils.py
mauritsvanrees: Done.

PR #421, opened 2023-07-12
Fix reading pypirc

mauritsvanrees:
Easy enough at first, but then the tests broke because my ~/.pypirc was found...  So some more fixes were needed.

### DIFF
--- a/zest/releaser/pypi.py
+++ b/zest/releaser/pypi.py
@@ -173,11 +173,15 @@ class PypiConfig(BaseConfig):
 
     def _read_configfile(self):
         """Read the PyPI config file and store it (when valid)."""
-        if not os.path.exists(self.config_filename):
+        config_filename = self.config_filename
+        if not os.path.exists(config_filename) and not os.path.isabs(config_filename):
+            # When filename is .pypirc, we look in ~/.pypirc
+            config_filename = os.path.join(os.path.expanduser("~"), config_filename)
+        if not os.path.exists(config_filename):
             self.config = None
             return
         self.config = ConfigParser(interpolation=None)
-        self.config.read(self.config_filename)
+        self.config.read(config_filename)
 
     def twine_repository(self):
         """Gets the repository from Twine environment variables."""
--- a/zest/releaser/pypi.py
+++ b/zest/releaser/pypi.py
@@ -278,12 +278,37 @@ class ZestReleaserConfig:
     hooks_filename = None
 
     def load_configs(self, pypirc_config_filename=DIST_CONFIG_FILE):
+        """Load configs from several files.
+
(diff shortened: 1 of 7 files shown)


zest/releaser/vcs.py
```diff
@@ -4,6 +4,7 @@
 
 import logging
 import os
+import pkg_resources
```
reinout, line 7: This is something that will bite us in the future: pkg_resources is going away, I believe :-)

zest/releaser/pypi.py
```diff
@@ -274,12 +278,37 @@ class ZestReleaserConfig:
...
+
+        * Most packages won't have this setting.
+        * If you make releases for lots of packages, you probably set this in
+          your global ~/.pypirc.
+        * A few individual packages will explicitly set this.
+          They will expect this to have the effect that the extra message is
+          added to commits, regardless of who makes a release.
+          So this individual package setting should win.
+        * Finally, pyproject.toml is newer than setup.cfg, so it makes sense
+          that this file has the last say.
```
reinout, line 302: Agreed

zest/releaser/pypi.py
```diff
@@ -274,12 +278,37 @@ class ZestReleaserConfig:
...
+        """
         setup_config = SetupConfig()
         pypi_config = PypiConfig(config_filename=pypirc_config_filename)
         pyproject_config = PyprojectTomlConfig()
         combined_config = {}
-        # overwrite any duplicate keys in the following order:
-        for config in [setup_config, pypi_config, pyproject_config]:
+        config_files = [pypi_config]
+        if self.use_package_config:
+            config_files.extend([setup_config, pyproject_config])
```
reinout, line 310: I'm wondering what this `use_package_config` setting does. It might need explaining in the `__init__()`.

..... Reinout reads the tests .....

Aha! this is only for during the test. `omit_package_config_in_test` would be clearer (default `False`, I think the negative term is clearer here).

zest/releaser/vcs.py
mauritsvanrees: True, but I copied this from `baserelease.py`.  And we are using `pkg_resources` in other places as well, for example for the hooks.

reinout: I'm also using it in lots of places myself, so I was a bit scared when I saw that it was probably going to be deprecated somewhere in the future :-)

Fine to leave it in here, for now, of course.

zest/releaser/pypi.py
mauritsvanrees: Done.  Clearer that way, yes.

reinout:
Looks fine!

Issue #420, opened 2023-07-12
Issues when trying 'build'


mauritsvanrees:
Using master here, after the merge of PR #415.

`build` seems to produce more warning output:

```
reading manifest template 'MANIFEST.in'
warning: no previously-included files matching '*.pyc' found anywhere in distribution
...
There were errors or warnings. Are you sure you want to continue? (y/N)? 
```

Workaround: [cleanup `MANIFEST.in`](https://github.com/mauritsvanrees/mauritstestpackage/commit/955a0847fe257a17150101762f7288459b46c8e4) in my test package.
This is unfortunate, but I don't think we can do anything about that.
cc @gforcada as we will likely see this a lot when releasing Plone packages as well.

More problematic, maybe not even related to `build`, is that I get this error when releasing:

```
ERROR: You must have a properly configured .pypirc file in your home dir to upload to a Python package index.
```

I very much *do* have this properly configured.  So somehow this is not recognised.  Probably an inadvertent change in the big PR from @elisallenens.  I am investigating.

reinout:
The `*.pyc` warning was what bugged @elisallenens and me in the big PR. That's where the "don't write bytecode" line in one of the test files came from.

iirc build would complain about missing `*.pyc` files that it is being told to ignore.

reinout:
I wondered a bit whether it was an interplay between "build" and "check-manifest" that we include as a recommended tool, btw.

mauritsvanrees:
Actually, we are already trying to [ignore that warning](https://github.com/zestsoftware/zest.releaser/blob/fb76b21ccaa2e051feb9c43dfc1b716eb974bc96/zest/releaser/utils.py#L637-L644).  Must be slightly different here.  I will see what I can do.

Apart from this problem, what has always irritated me:

* You have a manifest where you say "ignore this file".
* You create a distribution.  This obviously does *not* contain the file.
* You install the distribution.
* setuptools/distutils/pip complains that you are ignoring a file that is not included...

But that is something that should have been solved in core Python 15 years ago already.  Anyway....

reinout:
Ah! Look at `_project_builder_runner()` in `release.py`. That's a function that gets passed to the "build" mechanism. 
The `KNOWN_WARNINGS` should probably be used in there.

mauritsvanrees:
Yes, I already found it, and am completely replacing it. :-)

mauritsvanrees:
All fixed.

mauritsvanrees:
[...]

PR #423, opened 2023-07-12
Fixed detecting errors/warnings in output of the 'build' command.

mauritsvanrees:
Fixes the final part of issue #420.

### DIFF
--- a/zest/releaser/release.py
+++ b/zest/releaser/release.py
@@ -2,9 +2,6 @@
 
 from build import ProjectBuilder
 from colorama import Fore
-from subprocess import CalledProcessError
-from subprocess import check_output
-from subprocess import STDOUT
 from urllib import request
 from urllib.error import HTTPError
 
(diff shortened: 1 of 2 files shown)


reinout (approved):
Wow, that's quite some simplification! If it works for you, LGTM :-)

Ping @elisallenens as it simplifies his code a lot, which means there might be corner cases that have been missed.

mauritsvanrees:
I think we are good to go for an alpha release now.
Thanks for your work @elisallenens, and @reinout for all the reviews!

reinout:
Yeah, do the release.

elisallenens:
Oh, this is much better. I didn't really dare touch the show_interesting_lines because I wasn't sure just how it worked together with git and twine, but this is much cleaner than first setting a colour on stderr and then reading the colour.

mauritsvanrees:
I have released 9.0.0a1.

PR #424, opened 2023-07-13
Fix reading of text files when encoding is specified in the config

reinout:
On the other hand.... we can take it as a sign that the encoding config option isn't necessary anymore. So we could remove quite some lines of code. (Long live python3 btw!)


reinout:
#425 is a better solution, I think.

PR #425, opened 2023-07-13
Remove unneeded (and crashing) encoding config option

reinout:


reinout:
Better fix than #424

mauritsvanrees (approved):
LGTM.
I did not find any package on my system that uses this option.  I think I might have in the past, or that I had problems in at least one package, that I maybe simply fixed by changing non-ascii to ascii.

PR #427, opened 2023-07-14
Ignore error output when calling `build`.

mauritsvanrees:
We only need to look at the exit code to see if it worked. See https://github.com/zestsoftware/zest.releaser/pull/415#issuecomment-1635596826

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,9 @@ Changelog for zest.releaser
 9.0.0a2 (unreleased)
 --------------------
 
+- Ignore error output when calling `build`.
+  We only need to look at the exit code to see if it worked.
+
 - Removed ``encoding`` config option as nobody is using it anymore (using the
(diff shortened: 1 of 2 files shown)


CHANGES.rst
```diff
@@ -4,6 +4,9 @@ Changelog for zest.releaser
 9.0.0a2 (unreleased)
 --------------------
 
+- Ignore error output when calling `build`.
+  We only need to look at the exit code to see if it worked.
```
reinout, line 8:
```suggestion
- Ignore error output when calling `build`.
  We only need to look at the exit code to see if it worked.
  You can call zest.releaser with ``--verbose`` if you want 
  to see the possible warnings.
```

reinout (approved):

CHANGES.rst
mauritsvanrees: Ah nice, I did not realise this.  Checked, and working.

PR #428, opened 2023-07-14
Simplify longtest, which makes it support markdown

### LINKED ISSUE #363: Support Markdown file validation under longtest
Hi,
I am always used to the tool `longtest` to validate files that are included in setup.py. I am wondering if it is possible to validate Markdown files in the same way.

Thanks

reinout:
Fixes #363

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,7 +13,7 @@ Changelog for zest.releaser
   the result in the browser, without error handling. ``twine check`` should be
   used if you want a real hard check (``longtest --headless`` is
   deprecated). The advantage is that longtest now also renders markdown
-  correctly.
+  correctly.  This adds `readme_renderer[md]` as dependency.
   Fixes `issue 363 <https://github.com/zestsoftware/zest.releaser/issues/363>`_.
 
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,7 +6,7 @@ Changelog for zest.releaser
 
 - Ignore error output when calling `build`.
   We only need to look at the exit code to see if it worked.
-  You can call zest.releaser with ``--verbose`` if you want 
+  You can call zest.releaser with ``--verbose`` if you want
   to see the possible warnings.
 
 - Removed ``encoding`` config option as nobody is using it anymore (using the
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,13 @@ Changelog for zest.releaser
   we don't use python 2 anymore. Fixes `issue 391
   <https://github.com/zestsoftware/zest.releaser/issues/391>`_.
 
+- The ``longtest`` is now simpler. It runs readme_renderer and just displays
+  the result in the browser, without error handling. ``twine check`` should be
+  used if you want a real hard check (``longtest --headless`` is
+  deprecated). The advantage is that longtest now also renders markdown
+  correctly.
+  Fixes `issue 363 <https://github.com/zestsoftware/zest.releaser/issues/363>`_.
+
 
 9.0.0a1 (2023-07-13)
 --------------------
--- a/README.rst
+++ b/README.rst
@@ -164,9 +164,9 @@ Note: markdown files should use the "underline" style of headings, not the
 
 There are some additional tools:
 
-- **longtest**: small tool that renders a setup.py's long description
-  and opens it in a web browser. This assumes an installed docutils
-  (as it needs ``rst2html.py``).
+- **longtest**: small tool that renders the long description and opens it in a
+  web browser. Handy for debugging formatting issues locally before uploading
+  it to pypi.
 
 - **lasttagdiff**: small tool that shows the *diff* of the current
   branch with the last released tag.  Handy for checking whether all
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,14 +13,13 @@ authors = [
     {name = "Maurits van Rees", email = "maurits@vanrees.org"},
 ]
 dependencies = [
-    "setuptools",
+    "build >= 0.6.0.post1",  # 0.6.0 wasn't compatible with Python 3.7
     "colorama",
     "readme_renderer",
     "requests",
-    "twine >= 1.6.0",
-    "build >= 0.6.0.post1",  # 0.6.0 wasn't compatible with Python 3.7
-    "tomli; python_version<'3.11'",
     "setuptools >= 61.0.0",  # older versions can't read pyproject.toml configurations
+    "tomli; python_version<'3.11'",
+    "twine >= 1.6.0",
 ]
 requires-python = ">=3.8"
 classifiers = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,7 +15,7 @@ authors = [
 dependencies = [
     "build >= 0.6.0.post1",  # 0.6.0 wasn't compatible with Python 3.7
     "colorama",
-    "readme_renderer",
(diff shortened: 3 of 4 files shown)


CHANGES.rst
```diff
@@ -9,6 +9,13 @@ Changelog for zest.releaser
   we don't use python 2 anymore. Fixes `issue 391
   <https://github.com/zestsoftware/zest.releaser/issues/391>`_.
 
+- The ``longtest`` is now simpler. It runs readme_renderer and just displays
+  the result in the browser, without error handling. ``twine check`` should be
+  used if you want a real hard check (``longtest --headless`` is
+  deprecated). The advantage is that longtest now also renders markdown
+  correctly.
```
mauritsvanrees, line 16:
```suggestion
  correctly.  This adds `readme_renderer[md]` as dependency.
```

mauritsvanrees (changes requested):
Generally, this should be okay.
But I tried it and it failed:

```
(Pdb) print(html)
File "/Users/maurits/.pyenv/versions/3.11.4/lib/python3.11/importlib/metadata/__init__.py", line 563, in from_name
return next(cls.discover(name=name))
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
StopIteration

During handling of the above exception, another exception occurred:

File "<frozen runpy>", line 198, in _run_module_as_main
File "<frozen runpy>", line 88, in _run_code
File "/Users/maurits/community/zest.releaser/lib/python3.11/site-packages/readme_renderer/__main__.py", line 62, in <module>
main()
File "/Users/maurits/community/zest.releaser/lib/python3.11/site-packages/readme_renderer/__main__.py", line 26, in main
message = metadata(args.input)
^^^^^^^^^^^^^^^^^^^^
File "/Users/maurits/.pyenv/versions/3.11.4/lib/python3.11/importlib/metadata/__init__.py", line 998, in metadata
return Distribution.from_name(distribution_name).metadata
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/Users/maurits/.pyenv/versions/3.11.4/lib/python3.11/importlib/metadata/__init__.py", line 565, in from_name
raise PackageNotFoundError(name)
importlib.metadata.PackageNotFoundError: No package metadata was found for Products.PlonePAS
```

So we would somehow need to build the package before calling this.  Otherwise it will either fail, or it will show an outdated long description.  Maybe run `build`?

mauritsvanrees:
I tried it out in the [`Plone` package](https://github.com/plone/Plone), which uses Markdown, and there it worked.  This is in a checkout that already contains an egg-info directory, which makes it work.

And the good thing is, if I edit the readme and run `longtest`, the change gets picked up.
Aha, and when I run `longtest`, `Plone.egg-info/PKG-INFO` and the other files get updated.

So why does this not work in my clone of [`Products.PlonePAS`](https://github.com/plone/Products.PlonePAS).  I ran `python3.11 setup.py egg_info`, and then `longtest`, but that did not help, still showing the same traceback I pasted above.

Trying it with [`plone.memoize`](https://github.com/plone/plone.memoize) it works just fine, also when no egg-info directory exists at the start.

So: must be something specific in `Products.PlonePAS`.  The new longtest command fails, even though it *does* create a correct egg-info directory.  The old longtest command passed though, which is strange.

mauritsvanrees:
It fails when I try it with [`Zope`](https://github.com/zopefoundation/Zope) as well.  Working theory: it fails when the code is in a `src` directory.

I have recent `readme_renderer` 40.0.0.  I tried with the previous 37.3, but that gives an error in all packages:

```
(Pdb) print(html)
usage: __main__.py [-h] [-o OUTPUT] input
__main__.py: error: argument input: can't open 'plone.memoize': [Errno 2] No such file or directory: 'plone.memoize'
```

Ah, version 40 has this in the changelog: "Add CLI option to render package README."  So we should require this as minimum version.

reinout:
Yeah, the `src` dir seems to be the "problem". The egg-info isn't findable.

If I make a virtualenv, do a dev-install of plone.pas and add zest.releaser, then bin/longtest picks up the package just fine.

mauritsvanrees:
It seems to be [this PR](https://github.com/pypa/readme_renderer/pull/281) that changed their use of `pkg_resources` into `importlib.metadata`.
Trying it out in `plone.memoize` it works, as the distribution is found:

```
>>> from pprint import pprint
>>> import importlib.metadata
>>> pprint(importlib.metadata.packages_distributions())
...
 'plone': ['plone.memoize'],
```

In `Products.PlonePAS` the same command show no `Products` anywhere.
When I do `cd src`, and try again, it *does* show up.

So a workaround/solution could be to do `os.chdir("src")` if that exists.  Surprisingly this even works when no egg-info directory is there yet.
Could be something that `readme_render` should fix though.

mauritsvanrees:
I am creating an issue there, will let it link here.

reinout:
So it is more *importlib* that cannot find the package metadata where *pkg_resources* can.
Could it be a combination of a pkg_resources-style namespace package with the src dir?

mauritsvanrees:
Actually, I see that `pkg_resources` has the same problem:

```
>>> import pkg_resources
>>> pkg_resources.get_distribution("Products.PlonePAS")
  File "<stdin>", line 1, in <module>
  File "/Users/maurits/.pyenv/versions/3.11.4/lib/python3.11/site-packages/pkg_resources/__init__.py", line 478, in get_distribution
    dist = get_provider(dist)
           ^^^^^^^^^^^^^^^^^^
  File "/Users/maurits/.pyenv/versions/3.11.4/lib/python3.11/site-packages/pkg_resources/__init__.py", line 354, in get_provider
    return working_set.find(moduleOrReq) or require(str(moduleOrReq))[0]
                                            ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/maurits/.pyenv/versions/3.11.4/lib/python3.11/site-packages/pkg_resources/__init__.py", line 909, in require
    needed = self.resolve(parse_requirements(requirements))
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/maurits/.pyenv/versions/3.11.4/lib/python3.11/site-packages/pkg_resources/__init__.py", line 795, in resolve
    raise DistributionNotFound(req, requirers)
pkg_resources.DistributionNotFound: The 'Products.PlonePAS' distribution was not found and is required by the application
```

If you first change directory, before calling pkg_resources, it works though:

```
>>> import os
>>> os.chdir("src")
>>> import pkg_resources
>>> pkg_resources.get_distribution("Products.PlonePAS")
Products.PlonePAS 8.0.3.dev0 (/Users/maurits/community/plone-coredev/6.0/src/Products.PlonePAS/src)
```

reinout:
Works fine in my local Products.PlonePAS dir now.

mauritsvanrees:
This works now for me.

It *does* seem that this still only works for projects that have a `setup.py`.  In that case we get the name of the package by first calling `python setup.py egg_info` (when getting `vcs.name`).  With only a `setup.cfg` or `pyproject.toml`, we parse those files, which means there is no `egg_info` directory, so rendering the readme fails.

But this is no worse than what we had before.  And it now works with Markdown, which is what we were mostly after.  I will merge.  Thanks!

mauritsvanrees:
I have released [9.0.0a2](https://pypi.org/project/zest.releaser/9.0.0a2/) with the latest changes.

reinout:
yeah

PR #432, opened 2023-07-28
Check the exit code of a command, instead of relying on stderr

mauritsvanrees:
When a command fails, always ask the user to continue or not.  I hope with this we strike a good balance between showing error output, exit codes, and asking if the user wants to continue.

With the previous alpha release, the balance was not good for me.  The final `git push` used `stderr` to report that everything was fine, and then `postrelease` treated this as an error, and asked me if I wanted to continue.  Now we check the exit code.

@elisallenens @gforcada Since Reinout is unavailable for a while, can you review it, or simply try this branch out when releasing some packages?  I could also do another alpha release to more easily try it.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,10 +5,10 @@ Changelog for zest.releaser
 --------------------
 
 - When a command we call exits with a non-zero exit code, clearly state this in the output.
+  Ask the user if she wants to continue or not.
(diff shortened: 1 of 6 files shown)


elisallenens:
I tried it out with some badly configured packages and it gives errors where it should and no errors where it shouldn't, so it seems fine to me.

mauritsvanrees:
Thanks for reviewing!
I have released 9.0.0b1.

PR #439, opened 2023-11-25
Add changelog entries on tag commit

### LINKED ISSUE #218: Use the changelog to generate an better commit message (or annotated tag)
GitHub uses the commit information (or the tag annotation) to make the release notes (see for example, https://github.com/zestsoftware/zest.releaser/releases/tag/6.10) 

It would be nice to take from the corresponding section from the change log and use it.

gforcada:
Fixes #218 

My main use case here is `zest.releaser` used together with `zestreleaser.towncrier`.

Thus the changes here are almost a _noop_, as the new key added to `DATA` (`history_this_release`) is expected to be filled on `zestreleaser.towncrier`.

For plain `zest.releaser` use case, we might have to parse the change log file manually to extract that information, right? Or is the `history_last_release` a good candidate already? 🤔 

I was not 100% sure, so that's why I'm creating this PR as a draft.

I'm pushing the `zestreleaser.towncrier` changes in a bit, stay tuned 📻


reinout:
When I hear `towncrier` I cry " @mauritsvanrees " right away. I'm only using manually tweaked changelog entries, Maurits knows more about towncrier I believe.

gforcada:
@mauritsvanrees so, would you suggest that the key to use should be `tag-message` ? 👍🏾 I'm happy with any name, as long as we can get this somehow merged 😄

mauritsvanrees:
Yes, the `tag-message` key would seem to work best.
This would be done by adapting your PR https://github.com/collective/zestreleaser.towncrier/pull/26, which would mean no changes in `zest.releaser` itself are needed, if I see it correctly.

PR #440, opened 2023-12-19
Add `upload-pypi` option to disable upload to Pypi

leplatrem:
I know this PR is not ready to merge, but would like to hear your feedback before polishing docs and writing tests.

I'm using zest.releaser with an `after_checkout` that builds the UI with Node.

I want to publish the built wheels/archives from a Github Action, using the official action that supports trusted publishing. See https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/

For that, I want zest.releaser to build everything but without uploading to Pypi. 

I would run it with:

```
[zest.releaser]
create-wheel = yes
upload-pypi = no
```

and `release --no-input`

What do you think?

> Note: I could also run the build commands directly in the Github Action instead of using `release`

### DIFF
--- a/zest/releaser/pypi.py
+++ b/zest/releaser/pypi.py
@@ -467,6 +467,19 @@ class ZestReleaserConfig:
             return False
         return self.config.get("create-wheel", True)
 
+    def upload_pypi(self):
+        """Should we upload the package to Pypi?
+
+        [Configure this mode by adding a ``upload-pypi`` option::
+
+            [zest.releaser]
+            upload-pypi = no
+
+        The default when this option has not been set is True.
+
+        """
+        return self._get_boolean('zest.releaser', 'upload-pypi', default=True)
+
     def register_package(self):
         """Should we try to register this package with a package server?
 
--- a/zest/releaser/release.py
+++ b/zest/releaser/release.py
@@ -155,6 +155,9 @@ class Releaser(baserelease.Basereleaser):
                 self.data["tagworkingdir"],
             )
             builder.build("wheel", "./dist/")
+        if not self.pypiconfig.upload_pypi():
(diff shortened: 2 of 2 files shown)


mauritsvanrees:
You can already say:

```
[zest.releaser]
release = no
```

But then no distributions are made, and in your case you *do* want the distributions, so this is no option, right?  Just checking if I understand you correctly.

In that case, your solution sounds fine.

Just wondering if it would be better to turn this into a question.  That 
could be done in the [`_ask_upload` method](https://github.com/zestsoftware/zest.releaser/blob/f873540b03c3ae2ed717019e798ecb6b55514d10/zest/releaser/release.py#L216).  `default` would be True, like it is now, and with the new option you can set it to False.  Since it is a question, you can still override it during release, except when you call it with `--no-input` of course.
But then, before this question is asked, we call [`is_pypi_configured`](https://github.com/zestsoftware/zest.releaser/blob/f873540b03c3ae2ed717019e798ecb6b55514d10/zest/releaser/release.py#L161) and this gives an error if you do not have a pypi configuration.
So the way you have it now may be the best.

leplatrem:
Thank you for your detailed answer.

> But then no distributions are made, and in your case you do want the distributions, so this is no option, right? Just checking if I understand you correctly.

Yes, that was the idea.

I'm sorry but in the end I won't need this work, another solution was chosen to inject the built UI into the package, and we don't use the hook anymore. 
Shall I close it or do you want to keep it open in case someone wants to take over?

reinout:
Your idea *does* sound like something I could use in the future, so I'd like to get it in.
I'm merging it into a branch instead of into master, then I can do a bit of cleanup and a changelog and so without you needing to keep your branch open. Okay?

PR #445, opened 2024-02-06
Use extras instead of deps in tox.ini

mtelka:
The current specification of test dependencies via the `deps` keyword is effectively saying that in addition to both `test` and `recommended` extras there needs to be also `zest.releaser` installed.  This causes inconvenience for downstream packagers because it is creating some sort of circular dependency.

See also: https://peps.python.org/pep-0508/#extras

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -4,8 +4,8 @@ envlist =
 
 [testenv]
 usedevelop = true
-deps =
-    zest.releaser[test]
-    zest.releaser[recommended]
+extras =
+    test
+    recommended
 commands =
     zope-testrunner --test-path={toxinidir} -s zest.releaser --tests-pattern=^tests$ {posargs:-v -c}


icemac (approved):
I like the suggested change, using `extras` in `zopefoundation` all the time.

reinout:
Thanks, looks good. I'm not *that* experienced with tox yet, so I've learned something again :-)

I wonder though: how can this influence downstream packaging? `tox.ini` is just for running the tests within this project, it isn't the same as the list of dependencies you get when you install it. The tox.ini *does* work, without any circular problems. 

Or are you using `tox.ini` as something of an installation instruction? Then the way the config is handled is really a sort of documentation. I'm asking as it might help me write better tox.ini files.

mtelka:
> I wonder though: how can this influence downstream packaging? `tox.ini` is just for running the tests within this project, it isn't the same as the list of dependencies you get when you install it. The tox.ini _does_ work, without any circular problems.

`tox` itself by default works without any issue because it installs `zest.releaser` anyway.

> Or are you using `tox.ini` as something of an installation instruction? Then the way the config is handled is really a sort of documentation. I'm asking as it might help me write better tox.ini files.

I'm packaging Python projects for OpenIndiana and during packaging we run tests to make sure everything is okay.  We prefer `tox` if available.  To run tests we do not use the default tox configuration, because it tests in virtual environment.  Since our packages runs in real environment testing in virtual env would be of little added value for us.

To run in real (current) environment we use the [tox-current-env](https://pypi.org/project/tox-current-env/) plugin.  Running in current env means that we cannot leave test dependency resolution on `tox`, but we need to provide them via operating system packages.  To get the list of needed testing deps we simply ask `tox`.  With the previous `deps`-style deps we've got `zest.releaser` too.

Thank you for merging this.

mtelka:
FYI, [here](https://github.com/OpenIndiana/oi-userland/blob/ec5eec6eadf7f0264b07fdf5be83d4f8ade5b53e/components/python/zest.releaser/patches/01-test-extras.patch) is this change as a downstream patch.

reinout:
Thanks for the info, I didn't know about tox-current-env (which might be something I can use later on, btw).

I don't know if you need a release for this, but as zest.releaser makes it so easy I put out 9.1.3 :-)

mtelka:
> I don't know if you need a release for this, but as zest.releaser makes it so easy I put out 9.1.3 :-)

No. I do not need release for this change only.  Actually, it is better to NOT release with this change only because every release adds some work for us and since this adds no value for us (we already have the patch downstream), then it is definitely better to go without the release yet :-).

Thank you.

Issue #448, opened 2024-05-26
Support Hatchling as build backend


hcording:
Hi, are there currently plans to support `hatch` as build backend? Since pyproject.toml is supported (#415), I think it would be an important feature to have. I have the following minimal code base:

**pyproject.toml**
```toml
[build-system]
requires = ["hatchling"]
build-backend = "hatchling.build"

[project]
name = "test"
dynamic = ["version"]
requires-python = ">=3.10"
license = "MIT"
authors = [
    { name = "Name", email ="mail@example.com" }
]
description = "Desc"
readme = "README.md"
classifiers = [
    "Programming Language :: Python :: 3.10",
    "Programming Language :: Python :: 3.11",
    "Programming Language :: Python :: 3.12",
    "License :: OSI Approved :: MIT License",
    "Operating System :: OS Independent",
    "Development Status :: 1 - Planning",
    "Intended Audience :: Developers",
    "Topic :: Software Development :: Libraries :: Application Frameworks"
]
dependencies = [

]
keywords = ["kw"]

[project.urls]
"Homepage" = ""
"Bug Tracker" = ""

[tool.hatch.version]
path = "version.txt"
pattern = '^(?P<version>[^\n]+)'
```

**version.txt**
```
0.1.0
```
Running `fullrelease`, I get this output:

```shell
INFO: Starting prerelease.
Do you want to run check-manifest? (Y/n)? y
lists of files in version control and sdist match
Run pyroma on the package before tagging? (Y/n)? y
INFO: ------------------------------
INFO: Checking /media/mint/projects/test
INFO: Getting metadata for wheel...
INFO: Creating isolated environment: venv+pip...
INFO: Installing packages in isolated environment:
- hatchling
INFO: Getting build dependencies for wheel...
INFO: Getting metadata for wheel...
INFO: Found test
INFO: ------------------------------
INFO: Final rating: 10/10
INFO: Your cheese is so fresh most people think it's a cream: Mascarpone
INFO: ------------------------------
WARNING: No detectable version heading in the history file CHANGELOG
Changelog entries for version 0.1.0:


Enter version [0.1.0]: 
INFO: History file CHANGELOG updated.
INFO: The 'git diff':

diff --git a/CHANGELOG b/CHANGELOG
index e69de29..388fd1a 100644
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -0,0 +1,2 @@
+0.1.0 (2024-05-26)
+------------------
\ No newline at end of file


[...]

PR #449, opened 2024-06-16
Build distributions in an isolated environment.

mauritsvanrees:
Otherwise `build` cannot install packages needed for the build system, for example `hatchling`. Fixes issue #448.

I see a few test failures locally, which should only be small differences in the output during release.  Let's see if they are there on GitHub as well.

Code is taken from how [`build` does it itself](https://github.com/pypa/build/blob/main/src/build/util.py#L41-L50).


mauritsvanrees:
I found out what was going wrong in the tests.  This actually has nothing to do with the changes in this PR.  The [scheduled tests already fail](https://github.com/zestsoftware/zest.releaser/actions/runs/9534246157/job/26278447566).  It took a long time before I saw what the actual diff was.  Wwith `setuptools>=69.3` sdists get an underscore.  Instead of `tha.example-0.1.tar.gz` you get `tha_example-0.1.tar.gz`.

Okay, I have updated the tests.  See commit 934ffb7a09e7db999a4220c90e9320d505cda1ac for more details.

reinout (approved):
Looks good! Thanks for figuring out the underscore/point problem... Looks like some serious debugging :-)

PR #450, opened 2024-07-01
chore(docs): no more pkg_resources

gforcada:
The snippet above the fix uses `importlib.metadata` rather than `pkg_resources`.


reinout (approved):

reinout:
Good one!

PR #456, opened 2025-03-03
Pkgresources fix + python version update

reinout:


reinout:
@mauritsvanrees : I changed `from pkg_resources import parse_version` into `from packaging.version import parse as parse_version` instead of `from packaging import version` and later on doing `version.parse(...)`.

I saw the `from packaging import version` variant in some plone code you mentioned in the plone plip thingy. Personally, I think that's a bit dangerous as it is really easy to use `version` as a variable name. We actually had some `version` arguments in the `utils.py`, which would have shadowed the `version` import. Something to watch out for.

stevepiercy (changes requested):
Don't forget docs, readme, and dockerfile. Otherwise LGTM.

https://github.com/search?q=repo%3Azestsoftware%2Fzest.releaser+3.8&type=code

mauritsvanrees (commented):
Overall this is good, thank you both!

But I would prefer to keep Python 3.9 support, as that is what Plone 6.0 still supports.  I expect this should be possible by requiring `importlib_metadata` on 3.9, and then adjusting our code accordingly.

I am trying to look at that locally, but I already see errors/failures when running the tests with Python 3.10.  I am surprised that GitHub Actions does not fail in the same way.

For example the tests in `tests/pyproject-toml.txt` fail in the test of line 83 which calls `release.main()` and expect this at the end:

```
    Creating ...
    removing 'tha_example-0.1' ...
    Question: Upload to pypi (y/N)?
    Our reply: yes
    MOCK twine dispatch upload ... dist/tha.example-0.1.tar.gz
```

For me locally there are three problems here:

1. Instead of `removing 'tha_example-0.1' ...` I get `removing 'tha.example-0.1' ...`, so a dot instead of underscore.
2. Similarly, on the last line I don't get a dot, but an underscore: `tha_example-0.1.tar.gz`.
3. For me the answer ends with these lines, instead of a mock upload:

```
    Question: Do you want to continue without uploading? (y/N)? 
    Our reply: y
```

The first two are caused by the new setuptools release getting used, resulting in a different directory/file name.

The last one is because we are using an old `.pypirc` during testing, which should not be valid anymore when you would really use it to upload to PyPI.  I think we can stop using that in our tests.

I am looking into this, but I currently don't understand why GHA does not fail with the same errors.

reinout:
@mauritsvanrees : when releasing zest.releaser I got a deprecation warning from pypi, warning me that `zest.releaser-9.3.0-py3-none-any.whl` isn't normalized enough, it should be `zest_releaser-9.3....`, with an underscore.
I didn't know about that, so I assume it is a relatively recent change. I also noticed the underscore in a work project.

mauritsvanrees:
@reinout I have been getting that warning for several packages since last year already, also depending on who is doing the release, as it depends on what versions you have installed.  I think it goes fine with a reasonably recent setuptools.

Actually, it can be that I only saw this warning for uploaded source distributions.  IIRC `setuptools` has fixed that in setuptools 70 (which wreaked havoc in `zc.buildout`, but I fixed that last year).  Until recently, I think wheels still had a non-normalised file name.  I already wondered when that would start giving problems.  I always expected this would need some change in the `wheel` package.  But last week setuptools 75.8.1 was released and this changed the filename of the wheel.  (And a day or two later 75.8.2 was released, which gives a better chance of finding an installed distribution with both the original name and the canonical (normalised) name.).

So I guess PyPI has started issuing these warnings recently.  Ah, [here it is](https://github.com/pypi/warehouse/pull/17378).

So: if you use setuptools 75.8.1 or higher for generating the wheel (and the sdist), you should be fine.

PR #457, opened 2025-03-03
pkgresources fix, but restore Python 3.9 support and fix local errors

mauritsvanrees:
This builds on PR #456, but restores Python 3.9 support by using `importlib_resources` (only on that version), and fix local errors where I do not understand that they are not going wrong on GitHub Actions.

See [my comment](https://github.com/zestsoftware/zest.releaser/pull/456#pullrequestreview-2654313660) on that PR.


reinout (approved):
Looks good!

PR #459, opened 2025-03-04
pkg_resources again

reinout:


mauritsvanrees (approved):
Ah yes, that is better.  Thanks!

PR #460, opened 2025-03-04
Add packaging to our dependencies.

mauritsvanrees:
We were already pulling this in via another dependency.

Remove even more remaining `pkg_resources` usage.


reinout:
Huh? I missed one? 
Anyway, looks good :-)

reinout (approved):

PR #462, opened 2025-03-05
Requiring "wheel" package

### LINKED ISSUE #461: Wheel not being build anymore: importlib.metadata.distribution gotcha?
When releasing zest.releaser, I only got a .tgz, not a .whl. Weird.

There *is* one recent change related to this. In `pypi.py`, we check for the presence of the `wheel` library by using `from importlib.metadata import distribution` and then `distribution("wheel"). I haven't figured out if something is wrong here, but it is suspicious enough to create an issue here.

reinout:
Fixes #461

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,10 @@ Changelog for zest.releaser
 9.3.2 (unreleased)
 ------------------
 
-- Nothing changed yet.
+- Requiring the ``wheel`` package now as everybody (rightfully so) uses wheels
+  nowadays. It used to be an optional dependency beforehand, though often automatically
+  included through setuptools' vendored libraries.
+  [reinout]
 
 
 9.3.1 (2025-03-04)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,8 @@ Changelog for zest.releaser
 - Requiring the ``wheel`` package now as everybody (rightfully so) uses wheels
   nowadays. It used to be an optional dependency beforehand, though often automatically
   included through setuptools' vendored libraries.
+  You can switch off creation of wheels by setting the option ``create-wheel = false``.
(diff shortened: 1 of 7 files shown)


doc/source/uploading.rst
```diff
@@ -109,18 +109,12 @@ Since version 6.6.6 we use it in a way that should work with ``twine``
 Uploading wheels
 ----------------
 
-First, you should install the ``zest.releaser[recommended]`` extra, or
-run ``pip install wheel`` yourself next to ``zest.releaser``.  Then
-create or edit ``setup.cfg`` in your project (or globally in your
-``~/.pypirc``) and add this to create and upload a wheel to upload to
-PyPI::
+The ``wheel`` library is a dependency of zest.releaser (since 9.4.0), so it is always
+installed.  In the (rare) case where you do not want wheels, you can switch it of in our
```
mauritsvanrees, line 113: typo

```suggestion
installed.  In the (rare) case where you do not want wheels, you can switch it off in our
```

CHANGES.rst
```diff
@@ -1,10 +1,13 @@
...
 ===========================
 
-9.3.2 (unreleased)
+9.4.0 (unreleased)
 ------------------
 
-- Nothing changed yet.
+- Requiring the ``wheel`` package now as everybody (rightfully so) uses wheels
+  nowadays. It used to be an optional dependency beforehand, though often automatically
+  included through setuptools' vendored libraries.
```
mauritsvanrees, line 9: Let's mention the option here as well, for good measure:

```suggestion
  included through setuptools' vendored libraries.
  You can switch off creation of wheels by setting the option ``create-wheel = false``.
  See our `options documentation <https://zestreleaser.readthedocs.io/en/latest/options.html>`_.
```

mauritsvanrees (approved):
Makes sense, thanks.

I added two textual suggestions.

reinout:
Back in business. 9.4.0 has both .tgz and .whl again.

PR #463, opened 2025-03-07
Really build source dist and wheel in isolation.

mauritsvanrees:
Previously we passed along our own `PYTHONPATH`, but that could lead to an unintended `setuptools` version being used.
See [my comment](https://github.com/plone/meta/pull/268#issuecomment-2703442444) on a `plone.meta` PR.

Also, when I ran `tox -e py39` locally, I got a traceback:

```
...
  File "/Users/maurits/community/zest.releaser/zest/releaser/__init__.py", line 12, in <module>
    __version__ = version("zest.releaser")
  File "/Users/maurits/.pyenv/versions/3.9.21/lib/python3.9/importlib/metadata.py", line 569, in version
    return distribution(distribution_name).version
  File "/Users/maurits/.pyenv/versions/3.9.21/lib/python3.9/importlib/metadata.py", line 542, in distribution
    return Distribution.from_name(distribution_name)
  File "/Users/maurits/.pyenv/versions/3.9.21/lib/python3.9/importlib/metadata.py", line 196, in from_name
    raise PackageNotFoundError(name)
importlib.metadata.PackageNotFoundError: zest.releaser
```

So in the second commit, we try hard to determine our own installed version, but do not fail when this does not work.  Report the version as "unknown" then.  There is no sense in failing to start up just because Python packaging is having a rough day...  Or week...


mauritsvanrees:
Note to self: would it help on Python 3.9 to always use the extra `importlib-metadata` package instead of `importlib.metadata` from the standard lib?  Just an idea while I was brushing my teeth. :-)

CHANGES.rst
```diff
@@ -4,7 +4,14 @@ Changelog for zest.releaser
 9.4.1 (unreleased)
 ------------------
 
-- Nothing changed yet.
+- Try hard to determine our own installed version, but do not fail when this does not work.
```
reinout, line 7: Too much detail for a changelog entry? The three lines can be summarized as “Worked around python 3.9 version-reporting issue that prevented startup”?

reinout (approved):
Looks good!
Only one changelog comment. 

I saw the 3.9 local failure, too, btw. Ignored it as GitHub worked fine. Weird…

Hope everything works again now.

CHANGES.rst
mauritsvanrees: Agreed and done.

reinout:
Released as `9.5.0`.
(Though I did get an error about my branch not being up to date after releasing despite doing a `git pull` beforehand, so I'm looking into that).

I really like the "these are the two files that will be uploaded" message, btw!

reinout:
Screenshot 2025-03-08 at 21 56 57

We've been fullreleasing at exactly the same time with the same software version, so there were no merge conflicts :-)

PR #469, opened 2025-04-10
Use the SPDX format now for the license field

reinout:


mauritsvanrees (dismissed):
AFAIK, the classifier needs to stay the same.
https://pypi.org/classifiers/ does not have SPDX spelling, and `pyroma` complains:

```
$ bin/pyroma .
------------------------------
Checking .
Getting metadata for wheel...
Creating isolated environment: venv+pip...
Installing packages in isolated environment:
- setuptools
- wheel
Getting build dependencies for wheel...
Getting metadata for wheel...
Found zest.releaser
------------------------------
Some of your classifiers are not standard classifiers:
GPL-2.0-or-later
Your package does neither have a license field nor any license classifiers.
------------------------------
Final rating: 9/10
Cottage Cheese
------------------------------
```

So the classifier should stay the same.

Actually, if I revert the classifier change, pyroma fails with a big error.  The most important part is this:

```
configuration error: `project.license` must be valid exactly by one definition (2 matches found):

    - keys:
        'file': {type: string}
      required: ['file']
    - keys:
        'text': {type: string}
      required: ['text']
```

Strange that this output is so different in the two cases.  I guess a different code path gets taken.
Upstream issue: https://github.com/regebro/pyroma/issues/93

mauritsvanrees (approved):
Should be fine now.  Now pyroma is not completely happy, but at least it does not crash:

```
Your package does neither have a license field nor any license classifiers.
------------------------------
Final rating: 9/10
Cottage Cheese
```

Can you look at my latest changes?  Then it can be merged.

reinout:
*Removing* the license from the classifiers is a good one.
setuptools (also 77+!) complains bitterly if the old classifier is present:

    setuptools.errors.InvalidConfigError: License classifiers have been superseded by license expressions (see https://peps.python.org/pep-0639/). Please remove:
    
    License :: OSI Approved :: GNU General Public License (GPL)

A build fails, then. Weird that it complains about a classifier, in a sense.

reinout:
Yeah, releasing ourselves is now nicely warning-free again (=> `9.6.2` is out).

PR #480, opened 2025-11-18
pre commit update and pyupgrade

mauritsvanrees:
I thought adding `pyupgrade` would automatically modernise things like `"string {}".format(something)` into f-strings, but apparently not.  So I did that manually.

### DIFF
--- a/zest/releaser/baserelease.py
+++ b/zest/releaser/baserelease.py
@@ -295,7 +295,7 @@ class Basereleaser:
             if index == 0:
                 line = f"{prefix} {line}"
             else:
-                line = "{}  {}".format(" " * len(prefix), line)
+                line = f"{' ' * len(prefix)}  {line}"
             lines.append(line)
         lines.append("")
         self.data["history_lines"][insert:insert] = lines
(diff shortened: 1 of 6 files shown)


mauritsvanrees:
Yes, I am doing that.  I am passing `--py39-plus`.  Did I get some indentation wrong in the pre-commit config?  No, it does not look like that.

I think we are hit by this comment in your pyupgrade link:

> note: pyupgrade is intentionally timid and will not create an f-string if it would make the expression longer or if the substitution parameters are sufficiently complicated (as this can decrease readability).

Indeed some of the manual changes I did were a bit tricky:

```
-            print("Checking version bump for {} release.".format(self.data["release"]))
+            print(f"Checking version bump for {self.data['release']} release.")
```

In that case pyupgrade may not know which double quotes to change.  Either of these lines would work:

```
+            print(f"Checking version bump for {self.data['release']} release.")
+            print(f'Checking version bump for {self.data["release"]} release.')
```

Also, take this test file:

```
a = "hello"
b = 1
# 1
print('%s %s' % (a, b))
# 2
print("%s %d" % (a, b))
```

pyupgrade only changes the first one.  First to `print('{} {}'.format(a, b))` and in a second run to `print(f'{a} {b}')`.
It does not touch the line with `%d`, probably because pyupgrade is not entirely sure, or thinks it would get more ugly.

I guess we were just unlucky.

Issue #474, opened 2025-09-11
Release fails for indexes other than PyPI and TestPyPI


cocorigon:
Twine 6.2.0 removed support for the `--skip-existing` flag for indexes other than PyPI and TestPyPI ([changelog](https://github.com/pypa/twine/blob/main/docs/changelog.rst#deprecations-and-removals))

However, this flag is always set by zest.releaser when performing a release, and there is no way for the user to overwrite it ([source code](https://github.com/zestsoftware/zest.releaser/blob/cd7e8da456b4bc7a5e730be392d8cd6b46693523/zest/releaser/release.py#L289)).

I am encountering this issue with Azure packages when running `fullrelease --no-input`:
```txt
twine.exceptions.UnsupportedConfiguration: ("The configured repository 'https://pkgs.dev.azure.com/my_org/my_project/_packaging/my_project/pypi/upload' does not have support for the following features: --skip-existing and is an unsupported configuration", 'https://pkgs.dev.azure.com/my_org/my_project/_packaging/my_project/pypi/upload', '--skip-existing')
```

icemac:
I have the same problem when using [devPi](https://github.com/devpi/devpi).

1letter:
i use pypiserver and remove the option in https://github.com/zestsoftware/zest.releaser/blob/ac2a27a840d19c835b47c6af070942010898fb5a/zest/releaser/release.py#L287
```
elif twine_command == "upload":
   # twine_args += ("--skip-existing",)
   pass
```

PR #483, opened 2025-12-15
Comment out skip-existing argument in twine upload

1letter:
Twine 6.2.0 removed support for the "--skip-existing"


mauritsvanrees (approved):
Thanks!  As said in [my comment](https://github.com/zestsoftware/zest.releaser/issues/474#issuecomment-3636819919) on your issue, it would have been nice if we could keep this option for uploads to the real PyPI.  But in normal usage it should not be needed there either.  So I approve.

This is missing a changelog entry, but I will merge and then add it.

PR #485, opened 2026-02-10
Support hatch's dynamic version

### LINKED ISSUE #484: Support hatch dynamic version
If you are using a dynamic version with `hatch`, e.g.:

```toml
[tool.hatch.version]
path = "src/foo/bar/__init__.py"
```

then `zest.releaser` fails to find the package version:

```shell
$ uvx --from zest-releaser  bumpversion
Checking version bump for normal release.
CRITICAL: No version found.
```

This can be fixed providing:
```toml
[tool.zest-releaser]
python-file-with-version = "src/foo/bar/__init__.py"
```

It would be cool if the [`get_pyproject_toml_version`](https://github.com/zestsoftware/zest.releaser/blob/c9139fa5015d69ba34366ea9a15ce13e7e17da03/zest/releaser/vcs.py#L145-L151) could be updated to parse that value.

reinout:
Fixes #484

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,9 @@ Changelog for zest.releaser
 9.8.1 (unreleased)
 ------------------
 
-- Nothing changed yet.
+- Added suport for ``path`` in ``[tool.hatch.version]``: hatch uses it for dynamic
+  version files. If found, it is used as the value for ``python-file-with-version`` in
+  zest.releaser's config.
 
 
 9.8.0 (2025-12-15)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,7 +1,7 @@
 Changelog for zest.releaser
 ===========================
 
-9.8.1 (unreleased)
+9.9.0 (unreleased)
 ------------------
 
 - Added suport for ``path`` in ``[tool.hatch.version]``: hatch uses it for dynamic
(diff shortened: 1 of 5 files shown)


mauritsvanrees (commented):
That should work.

It would be nicer if no dummy `tool.zest-releaser` would be needed.  I did not try it, but something like this:

```
    def zest_releaser_config(self):
        if self.config is None:
            return None
        try:
            result = self.config["tool"]["zest-releaser"]
        except KeyError:
            logger.debug(
                f"No [tool.zest-releaser] section found in the {self.config_filename}"
            )
            result = {}
        hatch_file_with_version = self._file_with_version_from_hatch()
        if hatch_file_with_version:
            logger.debug("Found version file path in the hatch config")
            # Inject it as if it is a zest.releaser config option.
            result["python-file-with-version"] = hatch_file_with_version
        return result if result else None
```

mauritsvanrees (approved):
LGTM.  @ale-rt Can you try this please?

ale-rt (approved):
Lovely!
Thanks a lot for the quick patch!
I am just leaving two totally ignorable comments.

reinout:
I don't want to add the extra checks for when someone misconfigured zestreleaser+hatch with two different files-with-version. Extra code that's not really needed. If something goes wrong, you can run with `--verbose` and it'll tell you which file the version number came from and you'll spot the error.

At least, that's what I think :-)

Issue #486, opened 2026-05-19
Package incorrectly reported as NOT on PyPI


mauritsvanrees:
I tried releasing `collective.recipe.backup`, but got this:

```
INFO: This package does NOT exist yet on PyPI.
Upload to pypi (y/N)? 
```

This is definitely not true.

I think I see the problem.

* We are checking this url: https://pypi.org/simple/collective.recipe.backup
* This should be: https://pypi.org/simple/collective-recipe-backup/

It seems that you get redirected fine in the browser, but not on the command line (`curl`) or with `urllib` like we do.  I only found out while checking the Test PyPI, on which the package *really* does not exist.  Then with `curl -v` I saw some more:

* https://test.pypi.org/simple/collective.recipe.backup gets redirected to:
* https://test.pypi.org/simple/collective.recipe.backup/ with an extra slash at the end.  This gives a 404.  So I tried this new url on the real PyPI:
* https://pypi.org/simple/collective.recipe.backup/ gets redirected to:
* https://pypi.org/simple/collective-recipe-backup/

PR #487, opened 2026-05-19
Fix checking if a package is on PyPI.

mauritsvanrees:
We were for example checking this url: https://pypi.org/simple/collective.recipe.backup
This worked for years.  But now it should be: https://pypi.org/simple/collective-recipe-backup/

Fixes issue #486.

### DIFF
--- a/zest/releaser/tests/functional.py
+++ b/zest/releaser/tests/functional.py
@@ -1,8 +1,6 @@
 """Set up functional test fixtures"""
 
-from io import StringIO
-from urllib import request
-from urllib.error import HTTPError
+from packaging.utils import canonicalize_name
 from zest.releaser import choose
 from zest.releaser import utils
 from zest.releaser.baserelease import NOTHING_CHANGED_YET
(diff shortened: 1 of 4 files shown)


mauritsvanrees:
I need to fix the tests.

mauritsvanrees:
The build is green.

zest/releaser/tests/functional.py
```diff
@@ -1,15 +1,14 @@
 """Set up functional test fixtures"""
 
-from io import StringIO
-from urllib import request
-from urllib.error import HTTPError
+from packaging.utils import canonicalize_name
```
reinout, line 3: Handy to know that this one exists!

zest/releaser/tests/functional.py
```diff
@@ -35,38 +34,27 @@ def _exit(code=None):
 
     sys.exit = _exit
 
-    # Monkey patch urllib for pypi access mocking.
-    test.orig_urlopen = request.urlopen
+    # Monkey the requests library for PyPI access mocking.
```
reinout:
```suggestion
    # Monkey patch the requests library for PyPI access mocking.
```

reinout (approved):

mauritsvanrees:
Released in 9.9.1

